PR #14, opened 2026-05-24
fix: drift detector ignores blank nodes

rustforrecess:
## Summary

`onto_drift` currently reports a flood of spurious changes when comparing two snapshots of any ontology that uses anonymous restriction classes (e.g. the Pizza tutorial). Blank-node identifiers are freshly minted on every parse, so the same file diffed against itself shows:

- ~40 "added" blank-node IRIs
- ~40 "removed" blank-node IRIs
- a Cartesian product of confidence-scored "likely renames" between them

…drowning any real entity changes in noise.

This PR filters `_:`-prefixed IRIs out of the vocabulary set the drift detector compares. Blank nodes are anonymous structural artifacts, so comparing them across snapshots tells you nothing useful and breaks the report.

## Repro before the fix

```bash
# Load the Pizza tutorial (or any OWL with restrictions)
open-ontologies onto_drift \
  --version-a pizza.ttl \
  --version-b pizza.ttl       # same file both sides
# Expected: empty drift
# Actual: ~40 added, ~40 removed, dozens of "renames", drift_velocity ≠ 0
```

After the fix, same call returns empty `added` / `removed` / `likely_renames`.

## Change

7 lines in `src/drift.rs::extract_vocabulary`:

```rust
let is_named = |iri: &str| !iri.starts_with("_:");

// ...in the class-vocab loop:
for iri in parse_iris(&json, "c").into_iter().filter(|i| is_named(i)) { ... }

// ...in the property-vocab loop:
for iri in parse_iris(&json, "p").into_iter().filter(|i| is_named(i)) { ... }
```

## Tests

Existing `tests/drift_test.rs` cases all use named IRIs (`ex:Dog`, `ex:Cat`, etc.) and aren't affected. I considered adding a blank-node-specific regression test but the existing suite already covers the named-IRI cases comprehensively — adding `tests/drift_blank_node_test.rs` would just verify the filter's literal behavior. Happy to add one if you'd prefer.

## Checklist

- [x] Inspected `tests/drift_test.rs` — existing cases use only named IRIs; the filter can't affect them
- [x] `cargo build` — passes (verified locally on Ubuntu via WSL)
- [x] `cargo test` — all crates green
- [x] `cargo clippy -- -D warnings` — no lints
- [x] `cargo audit` — only the pre-existing allowed warning
- [x] CHANGELOG.md updated under `[Unreleased] / Fixed`
- [x] Single focused change

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ All notable changes to Open Ontologies are documented here.
 - **`SchemaIntrospector::sql_to_xsd` extended** to handle DuckDB-native types (HUGEINT, U{TINY,SMALL,}INT, DOUBLE, parameterised DECIMAL/VARCHAR, DATETIME, UUID, TIME).
 - New tests: `tests/sqlsource_test.rs` (driver detection, no features required) and `tests/duckdb_test.rs` (introspection + query → row extraction, gated by the `duckdb` feature).
 
+### Fixed
+- **`onto_drift` ignores blank nodes**. Pizza-style ontologies (and any OWL with restriction classes) use anonymous blank-node restriction classes that get freshly reminted on every parse. Two snapshots of the same file would show ~40 added + ~40 removed bnodes plus a Cartesian product of confidence-scored "renames" between them, drowning real entity changes in noise. The vocabulary extractor now filters `_:`-prefixed IRIs from both class- and property-gather loops.
+
 ### Documentation
 - `docs/data-pipeline.md` rewritten to cover both file-based and SQL-based ingest paths, the supported connection-string forms, federation examples (Parquet on S3 + Postgres scanner + remote CSV in one query), and a build matrix for the new feature flags.
 - `SKILL.md`, `skills/ontology-engineering/SKILL.md`, `skills/ontology-engineer.md`, and `CLAUDE.md` Tool Reference tables expanded to cover the SQL backbone tools and previously-missing tools (`onto_status`, `onto_marketplace`, `onto_unload`, `onto_recompile`, `onto_cache_status`, `onto_cache_list`, `onto_cache_remove`, `onto_repo_list`, `onto_repo_load`, `onto_embed`, `onto_search`, `onto_similarity`, `onto_dl_explain`, `onto_dl_check`, `onto_import_schema`, `onto_sql_ingest`).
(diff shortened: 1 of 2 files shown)


fabio-rovai (approved):
Thanks for the clean diagnosis and minimal fix — this is exactly the right shape for the bug.

I traced the chain end-to-end to confirm: `GraphStore::sparql_select` uses Oxigraph's `term.to_string()` which serialises blank nodes as `_:bN` (no angle brackets), and `parse_iris` only strips `<` and `>`, so blank-node IRIs survive into the vocabulary map unchanged. Filtering on `_:` at the two `extract_vocabulary` call sites is the correct fix. The Pizza repro is accurate — `owl:Restriction` axioms are blank-node-backed by design and would generate exactly the noise you describe.

One ask before merge — please add the regression test you offered. I'd suggest making it assert the **observable bug** rather than the filter's mechanics:

```rust
// tests/drift_blank_node_test.rs
// Load a small ontology with at least one owl:Restriction into two
// snapshots, then diff. With the fix, all three result sets are empty;
// without the fix, the test fails loudly with phantom add/remove/rename
// entries.

#[test]
fn drift_of_same_ontology_ignores_blank_node_restrictions() {
    let ttl = r#"
        @prefix owl: <http://www.w3.org/2002/07/owl#> .
        @prefix ex:  <http://example.org/> .
        ex:Pizza a owl:Class .
        ex:Topping a owl:Class .
        ex:hasTopping a owl:ObjectProperty .
        ex:MeatyPizza a owl:Class ;
            rdfs:subClassOf [ a owl:Restriction ;
                              owl:onProperty ex:hasTopping ;
                              owl:someValuesFrom ex:Topping ] .
    "#;
    // load into two GraphStores, run DriftDetector, assert added/removed/likely_renames empty
}
```

The reason to assert *behaviour* not *mechanism*: a future refactor (e.g. swapping the SPARQL-based extraction for a direct Oxigraph iterator API) could silently drop the filter and reintroduce the noise — only a behaviour-level test catches that class of regression.

One minor design note, not blocking: the filter could live inside `parse_iris` itself since that helper is only used at these two sites and blank nodes never make sense as named-vocabulary IRIs anywhere in this codebase. Either placement is defensible; happy with the current explicit-at-call-site version.

Approving on the fix itself. Will merge once the regression test lands.

fabio-rovai:
Merged — thanks again for the clean diagnosis and minimal fix. I went ahead and landed the regression test as a follow-up commit on `main` (88accf5) so it didn't have to round-trip through you; kept the author credit for the fix on your commit. Asserts `detect(x, x)` returns empty on an OWL ontology with `owl:Restriction` blank-node classes, plus checks the non-trivial diff case doesn't leak `_:` IRIs. Verified failing-then-passing across `origin/main` pre-merge and post-merge.

Appreciate the contribution — looking forward to more.

PR #58, opened 2026-06-05
Add native Windows build support

ziozzang:
## Summary
- add Windows-specific fallbacks so the Rust engine builds on native Windows without breaking Unix paths
- update the Tauri Studio launcher and packaging flow to find Windows binaries and avoid Unix-only shell commands
- document the verified Windows toolchain and build steps, and extend CI/release workflows with Windows targets

## Verification
- `cargo test --lib`
- `cargo build --release --bin open-ontologies`
- `npm run tauri build`


ziozzang:
tested codex desktop on windows 11 with MCP, and works well. :)

https://github.com/ziozzang/open-ontologies/releases/tag/tag-windows

fabio-rovai (changes requested):
Thanks for this — the Windows support is well put together. The production code is correctly platform-gated throughout: the `socket_windows.rs` stub matches the real `serve(&str, Arc<GraphStore>)` signature, the `ServeUnix` CLI variant splits cleanly into `#[cfg(unix)]`/`#[cfg(windows)]` with a graceful dispatch arm, and the Studio runtime changes in `engine.rs`/`chat.rs` replace the Unix-only `sh`/`lsof`/Homebrew logic with proper cfg-gated `netstat`/`taskkill` equivalents. Nicely done.

One blocker before merge: the **`build (windows-latest)` CI job is failing to compile**.

Root cause is `tests/socket_test.rs`. It's an entirely Unix-specific test (it imports `tokio::net::UnixStream` and exercises the domain-socket adapter), but unlike the production code it never got a platform gate, so `cargo test` on Windows fails with:

```
error[E0432]: unresolved import `tokio::net::UnixStream`
error: could not compile `open-ontologies` (test "socket_test")
```

One-line fix — add an inner attribute at the very top of `tests/socket_test.rs`:

```rust
#![cfg(unix)]
```

That gates the whole file out on Windows and should turn the CI green.

Minor, non-blocking: a good chunk of the diff in `config.rs`/`main.rs`/`lib.rs` is rustfmt reformatting unrelated to Windows. No need to revert it here, but keeping formatting-only changes in a separate commit makes future reviews easier.

Once Windows CI is green I'm happy to merge.

fabio-rovai (approved):
All green on both platforms now — merging. Thank you for this contribution, Jioh! Native Windows support is a big step for the project, and the platform gating across the socket layer, CLI, and Studio runtime was cleanly done. Interesting side effect: once the test suite could compile on Windows, it surfaced a real startup stack-overflow in debug builds (Windows main threads get 1 MiB vs 8 MiB on Unix), which is now fixed on this branch too. Great to have the project running natively on Windows 11.

PR #64, opened 2026-07-02
Studio: multilingual label filters, 3D graph view toggle, tree connec…

ma-be:
## Summary

- **Tree/Graph toggle** — new button pair in the Studio toolbar switches the main panel between the virtualized tree and the 3D force-directed graph canvas; both views share the same node-select callback so Inspector and Lineage stay in sync.
- **Multilingual label filter (Graph)** — `GraphCanvas` detects all BCP-47 language tags in `rdfs:label` values and renders a language chip bar overlay. Switching locale relabels nodes in-place (canvas texture repaint + `needsUpdate=true`, link endpoints re-resolved from string IDs) without touching force-simulation positions. Fallback chain: preferred → `en` → untagged → first available → URI local name.
- **Multilingual label filter (Tree)** — `TreeView` gains the same chip bar. Two-pass load collects all label variants into `labelMapRef`; language switch walks and relabels existing React state with no SPARQL re-query. `nodeMapRef` updated so breadcrumbs and connection chips reflect the selected locale.
- **Language badges + filter in Property Inspector** — parses lang tags from both engine-native (`"value"@lang`) and SPARQL JSON (`xml:lang`) formats. Badge per literal row; chip bar filters to one locale (URIs always shown). SPARQL patterns include lang tag to avoid clobbering sibling translations. Add form gains optional lang tag input.
- **Tree connector bug fix** — ancestor-line loop ran `lvl < indent`, placing a full-height vertical at the same column as the node's own connector when the direct parent was not last, making last-children show T instead of L. Fixed by stopping at `lvl < indent - 1`.

## Test plan

- [ ] Open Studio with a multilingual ontology (e.g. DCAT from marketplace)
- [ ] Confirm Tree/Graph toggle switches views; Inspector opens on node click in both
- [ ] In Graph view: language chip bar appears; switching locale relabels nodes; edges stay connected
- [ ] In Tree view: language chip bar appears in header; switching locale relabels all nodes and breadcrumbs
- [ ] In Inspector: literal rows show language badges; chip bar filters correctly; editing preserves other locales
- [ ] In Tree view: last children always show L-shaped connector, not T


fabio-rovai (changes requested):
Nice feature set — the multilingual handling and view toggle are welcome, and the connector investigation was in the right place. Requesting changes for four confirmed issues (each verified against the branch source, with worked triggers); the fixes are mostly small and local.

## Confirmed bugs (most severe first)

1. **`studio/src/components/PropertyInspector.tsx` (~line 164) — edits can be committed onto the wrong triple and persisted.** `saveEdit`/`deleteProp` index into `visibleProperties[idx]`, but `editingIdx` is never reset when `langFilter` changes (only on node change, save, Escape, +Add). If the filter changes while an edit is open, `visibleProperties` reshuffles and Enter runs DELETE/INSERT against whatever property now sits at that index — wrong data written via `saveGraphToFile`. Two live triggers: (a) the auto-reset effect (~line 113) flips the filter to `all` when the filtered language disappears — e.g. deleting the last literal of that language via the row × button while another row's edit input is open; (b) a direct chip click while editing in Tauri's WKWebView, where buttons don't steal focus so the input never blurs. Fix: reset `editingIdx` (and `addingNew`) whenever `langFilter` changes, or key the edit to the property identity (predicate+value) instead of the row index.

2. **`GraphCanvas.tsx` (~line 205) and `TreeView.tsx` (~line 239) — stale `langFilter` closure silently reverts the language switch.** `loadGraph` (deps `[showToast, refreshStats]`) and `loadTree` (deps `[refreshStats]`) both read `langFilter` but omit it from their dependency arrays, and both are published as `window.__refreshGraph` and re-run after every mutation (add class, delete node, inspector edit, chat op, undo). Switch to `cs`, then perform any mutation: the reload rebuilds every label with the mount-time `'en'` while the `cs` chip stays highlighted, and the `[langFilter]` relabel effect doesn't re-fire because the state didn't change. This breaks the PR's headline feature on the first mutation after a language switch. Fix: add `langFilter` to the deps (or read it through a ref).

3. **`TreeView.tsx` (~line 672) — the connector fix is still off by one, just one level deeper.** The new `lvl < indent - 1` bound correctly kills the phantom T at the node's own column, but the surviving loop gates column `lvl * INDENT_W + 16` on `parentIsLast[lvl]`, while the ancestor whose continuation line lives at that column is the one at `lvl + 1` (a node at indent k draws its elbow and sibling line at column k−1). Worked example: roots `[classes (not last), properties (last)]`; under `properties`, child A (not last, expanded) then B (last). A's grandchildren have `parentIsLast = [true, false]`; the loop checks `!parentIsLast[0]` → draws nothing at column 0, so the vertical connecting A to B disappears across A's subtree. The converse case draws a floating line under an L. Fix: gate column `lvl` on `!parentIsLast[lvl + 1]` — or compute one glyph per depth level in a single pass so each column has exactly one owner by construction.

4. **`studio/package-lock.json` — ~230 lines of unexplained dependency churn** (all of `@babel/*` → 7.29.7, all `@esbuild/*` 0.27.4 → 0.28.1, lodash, browserslist, …) with a zero-line `studio/package.json` diff — an incidental `npm update` riding along in a UI PR. Please revert or split it out; a build or supply-chain issue from those bumps would be untraceable to this PR.

## Smaller items

- RDF term parsing is copy-pasted: `parseSparqlResults` + `pickLabel` duplicated verbatim in GraphCanvas and TreeView, `shortUri` in three files, and the literal/lang-tag regex in four places. One `studio/src/lib/sparql.ts` fixes all of it — `lib/` already exists.
- Each view keeps its own independent `langFilter` state, so setting the Tree to `cs` and toggling to Graph shows `en` — which contradicts the "views stay in sync" framing. A shared context plus one `<LangChips>` component would also fix TreeView's chip bar hardcoding hex colours (`#89b4fa`/`#1e1e2e`) where the other three copies use the theme variables.
- The locale-switch relabel over-works: GraphCanvas re-clones every link on each chip click (3d-force-graph treats them as new data and reheats the simulation, so the graph jiggles) and repaints/re-uploads every node texture even when the picked label didn't change; TreeView spread-clones the entire tree, giving every node new identity and defeating all downstream memoization. Skipping unchanged labels and using structural sharing makes chip clicks near-free.
- `nodesRef` in GraphCanvas is written but never read — delete it.
- `saveEdit`'s `newPart` inlines the escape logic that the PR's own `literalPart` helper (five lines up) exists to centralize; reuse it so DELETE and INSERT can't drift.

Happy to re-review once the two closure fixes and the `editingIdx` reset land — those three are small diffs.

fabio-rovai:
Five weeks with no response. That is on me. Your CI never ran either: fork PRs from first-time contributors park as `action_required` until a maintainer approves the workflow, and nobody did.

Rebased onto current main. One conflict, a CHANGELOG ordering clash, resolved by keeping both entries. The Rust side is green: `cargo test` gives 531 passed, 0 failed.

## What I still cannot verify

CI here is cargo only. Nothing in `ci.yml` builds, type-checks or tests `studio/`, so a green tick on this PR says nothing about the code it actually changes. Your diff is `GraphCanvas.tsx`, `Layout.tsx` and `PropertyInspector`, and none of it has been compiled by anything.

That is a gap in the repo, not a criticism of the PR. But it means before I merge I need one of:

- confirmation you have run the Studio build and exercised the tree/graph toggle and the language chips locally, ideally with a screenshot or a short clip of the locale switch, or
- your agreement that I add a `studio` job to CI (install, `tsc --noEmit`, build) as a prerequisite, which I would rather do anyway and which #85 has just made cheaper by establishing the multi-job pattern.

## On the change itself

The part I want to understand better before merging is the in-place relabel in `GraphCanvas`. Repainting the canvas texture with `needsUpdate = true` and re-resolving link endpoints from string IDs, specifically so that force-simulation positions survive a locale switch, is the right instinct: rebuilding the graph would scatter the layout and lose the user's mental map. What I cannot tell from the diff is whether the texture repaint is bounded by visible nodes or runs across the whole set, which matters on the larger ontologies in the marketplace.

The fallback chain (preferred, `en`, untagged, first available, URI local name) matches how `onto_align` resolves BCP-47 tags elsewhere in the codebase, so that is consistent and I have no notes on it.

Tell me which of the two verification routes you prefer. If you have gone quiet on this, say so and I will take it forward myself with your authorship preserved, which is a reasonable outcome given how long you waited.

fabio-rovai:
Apologies for the silence, and thank you for this.

Mechanically: the branch conflicts with main, which has moved a long way since 9 August, and CI has never run on it. `gh pr checks` reports no checks at all on `feature/studio-multilingual-graph-view`, so the Studio typecheck leg has not seen any of it. A rebase plus a green typecheck is what this needs before I can review the diff properly.

On the substance, the multilingual work is aimed at the right thing and lines up with where the engine already is. `onto_embed` runs a multilingual model by default and `onto_align` parses the BCP-47 tag off each label, so the store genuinely holds labels in several languages and Studio has been the one place that could not show them. The fallback chain you describe, preferred then `en` then untagged then first available then the URI local name, matches how the engine resolves them, which is what I would want: two components disagreeing about which label is "the" label is worse than either rule on its own.

Two things I will be looking at when it is rebased. First, relabelling in place without disturbing force-simulation positions is the correct call and also the easiest thing to regress, so I would like a test or at least a described manual check that switching locale twice returns the graph to its original layout. Second, the tree and the graph now each hold their own label map (`labelMapRef`, `nodeMapRef`); if those two can drift, the same node can carry different labels in the two views, and I would rather one of them owned the map.

Rebase when you have a moment and I will review it properly this time.

fabio-rovai:
I have just approved the workflow run on this branch, so it is executing now against 5cbbe31. That is the first time CI has ever run on this PR. Worth saying plainly, because until a minute ago the absence of checks here was indistinguishable from passing checks, and neither of us had a signal.

What stands between this and a merge is no longer a product question. The branch is conflicting against main, which has moved a fair distance since 9 August. And whatever the run tells us, it is telling us about a fortnight-old base, so a rebase is needed before the result means anything.

Both design decisions on the sibling PR were answered on 21 August and neither of them reshapes what you wrote here. The label work stands as designed: your fallback chain matching the engine's is the right call, since two components disagreeing about which label is the label is worse than either rule alone.

The review asks from that pass, still open, are that relabelling in place should not disturb force-simulation positions, so a check that switching locale twice restores the original layout, and that `labelMapRef` and `nodeMapRef` each hold their own map and can drift, so the tree and the graph can show different labels for the same node.

Push a rebase and I will approve the run again straight away.

PR #65, opened 2026-07-04
CLI: daemon mode, human-readable output, marketplace proxy

ma-be:
## Summary

- **Daemon mode** — `open-ontologies daemon start/stop/status` launches `serve-http` as a
  detached background process and writes its PID + URL to `~/.open-ontologies/daemon.json`.
  All 24 CLI commands that touch the graph store automatically detect a live daemon and proxy
  their request to it via `/api/batch` — no per-command flags needed. Use `--no-connect` to
  force local execution. New modules: `src/daemon.rs` (PID file + process control) and
  `src/connect.rs` (HTTP proxy client). No new Cargo dependencies.

- **Marketplace via daemon** — `marketplace list` and `marketplace install --id <id>` are now
  routable through the daemon's `/api/batch` endpoint. Installing an ontology with a daemon
  running loads it into the daemon's persistent in-memory store, so subsequent `stats`, `query`,
  and `reason` calls in any other process see the installed triples. Implemented via a new
  `exec_marketplace` async handler in `BatchRunner`.

- **Human-readable output by default** — CLI commands now print plain text (e.g. `stats` →
  key/value block, `load` → "Loaded N triples from …"). Pass `--json` for the previous JSON
  output or `--pretty` for indented JSON. A `OnceLock<bool>` global (`JSON_MODE`) is set once
  at startup; daemon proxy output strips the batch envelope and renders identically to local
  output. Human-readable rendering lives in the new `src/output.rs` module (stats tables,
  SPARQL result tables, marketplace lists, lint/enforce issue lists, version history, and a
  pretty-JSON fallback).

## Test plan

- [ ] `open-ontologies daemon start` → check `~/.open-ontologies/daemon.json` created with valid PID + URL
- [ ] Run `open-ontologies load <file>`, `stats`, `query` without flags — confirm human-readable output; repeat with `--json` and `--pretty`
- [ ] With daemon running, run `open-ontologies load <file>` in a second terminal — confirm it proxies to daemon (triples visible via `stats`)
- [ ] `open-ontologies marketplace list` with daemon running — confirm results are proxied and rendered correctly
- [ ] `open-ontologies marketplace install --id <id>` with daemon — confirm triples land in daemon store
- [ ] `open-ontologies daemon stop` — confirm process terminates and `daemon.json` is removed
- [ ] `open-ontologies --no-connect load <file>` with daemon running — confirm local execution, not proxied
- [ ] Stale `daemon.json` (daemon killed externally) — confirm next CLI command removes the stale file cleanly

### DIFF
--- a/src/batch.rs
+++ b/src/batch.rs
@@ -167,7 +167,11 @@ impl BatchRunner {
             None => return json!({"error": "validate requires a file path"}),
         };
         match GraphStore::validate_file(input) {
-            Ok(count) => json!({"ok": true, "triples": count}),
+            Ok(counts) => json!({
+                "ok": true,
+                "triples": counts.triples,
+                "statements": counts.statements,
+            }),
             Err(e) => json!({"error": e.to_string()}),
         }
     }
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -7,6 +7,15 @@ use oxigraph::model::*;
 use oxigraph::sparql::{QueryResults, SparqlEvaluator};
 use oxigraph::store::Store;
 
+/// What a parse produced. `statements` counts parser events; `triples` counts
+/// distinct triples, which is the size of the resulting graph. They differ
+/// whenever the source repeats a statement.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct ValidationCounts {
+    pub statements: usize,
+    pub triples: usize,
+}
+
 /// Optional HTTP authentication for remote SPARQL endpoints.
 ///
 /// Enterprise triple stores gate their SPARQL Protocol endpoints behind auth:
@@ -110,10 +119,14 @@ impl GraphStore {
         let quads: Vec<_> = parser
             .for_reader(reader)
             .collect::<Result<_, _>>()?;
+        // Report triples actually added, not parse events. The store is a set, so
+        // re-inserting a statement it already holds changes nothing and must not be
+        // counted as a load.
+        let before = store.len().unwrap_or(0);
         for quad in &quads {
             store.insert(quad)?;
         }
-        Ok(quads.len())
+        Ok(store.len().unwrap_or(before).saturating_sub(before))
     }
 
     /// Load RDF content in a specified format (Turtle, RDF/XML, etc.)
@@ -133,10 +146,14 @@ impl GraphStore {
         let quads: Vec<_> = parser
             .for_reader(reader)
             .collect::<Result<_, _>>()?;
+        // Report triples actually added, not parse events. The store is a set, so
+        // re-inserting a statement it already holds changes nothing and must not be
+        // counted as a load.
+        let before = store.len().unwrap_or(0);
         for quad in &quads {
             store.insert(quad)?;
         }
-        Ok(quads.len())
+        Ok(store.len().unwrap_or(before).saturating_sub(before))
     }
 
     pub fn load_file(&self, path: &str) -> anyhow::Result<usize> {
@@ -160,10 +177,14 @@ impl GraphStore {
         let quads: Vec<_> = parser
             .for_reader(reader)
             .collect::<Result<_, _>>()?;
+        // Report triples actually added, not parse events. The store is a set, so
+        // re-inserting a statement it already holds changes nothing and must not be
+        // counted as a load.
+        let before = store.len().unwrap_or(0);
         for quad in &quads {
             store.insert(quad)?;
         }
-        Ok(quads.len())
+        Ok(store.len().unwrap_or(before).saturating_sub(before))
     }
 
     pub fn save_file(&self, path: &str, format: &str) -> anyhow::Result<()> {
@@ -172,28 +193,42 @@ impl GraphStore {
         Ok(())
     }
 
-    pub fn validate_turtle(ttl: &str) -> anyhow::Result<usize> {
+    pub fn validate_turtle(ttl: &str) -> anyhow::Result<ValidationCounts> {
         let reader = Cursor::new(ttl.as_bytes());
         let parser = RdfParser::from_format(RdfFormat::Turtle).for_reader(reader);
-        let mut count = 0;
-        for quad in parser {
-            quad?;
-            count += 1;
-        }
-        Ok(count)
+        Self::count_parsed(parser)
     }
 
-    pub fn validate_file(path: &str) -> anyhow::Result<usize> {
+    pub fn validate_file(path: &str) -> anyhow::Result<ValidationCounts> {
         let content = std::fs::read_to_string(path)?;
         let format = Self::detect_format_sniffed(path, &content);
         let reader = Cursor::new(content.as_bytes());
         let parser = RdfParser::from_format(format).for_reader(reader);
-        let mut count = 0;
+        Self::count_parsed(parser)
+    }
+
+    /// Count what a parser produced, distinguishing statements from triples.
+    ///
+    /// An RDF graph is a set, so a statement repeated in the source contributes
+    /// one triple and not two. Reporting the parse-event count as a triple count
+    /// overstates any generated document that repeats a statement, which real
+    /// serialisers do constantly: emitting `?lib a :Library` once per record is
+    /// ordinary practice and inflated one 16.7 MB file by 6.7 per cent.
+    fn count_parsed<I>(parser: I) -> anyhow::Result<ValidationCounts>
+    where
+        I: IntoIterator<Item = Result<Quad, oxigraph::io::RdfParseError>>,
+    {
+        let mut statements = 0usize;
+        let mut seen = std::collections::HashSet::new();
         for quad in parser {
-            quad?;
-            count += 1;
+            let quad = quad?;
+            statements += 1;
+            seen.insert(quad);
         }
-        Ok(count)
+        Ok(ValidationCounts {
+            statements,
+            triples: seen.len(),
+        })
     }
 
     pub fn sparql_select(&self, query: &str) -> anyhow::Result<String> {
--- a/src/main.rs
+++ b/src/main.rs
@@ -1436,8 +1436,12 @@ async fn async_main() -> anyhow::Result<()> {
                 GraphStore::validate_file(&input)
             };
             match result {
-                Ok(count) => output_json(
-                    &serde_json::json!({"ok": true, "triples": count}),
+                Ok(counts) => output_json(
+                    &serde_json::json!({
+                        "ok": true,
+                        "triples": counts.triples,
+                        "statements": counts.statements,
+                    }),
                     cli.pretty,
                 ),
                 Err(e) => {
--- a/src/ontology.rs
+++ b/src/ontology.rs
@@ -15,9 +15,10 @@ impl OntologyService {
     /// Validate RDF syntax. Returns a JSON report (never errors on bad input).
     pub fn validate_string(content: &str) -> anyhow::Result<String> {
         match GraphStore::validate_turtle(content) {
-            Ok(count) => Ok(serde_json::json!({
+            Ok(counts) => Ok(serde_json::json!({
                 "valid": true,
-                "triple_count": count,
+                "triple_count": counts.triples,
+                "statement_count": counts.statements,
                 "errors": []
             })
             .to_string()),
@@ -33,10 +34,11 @@ impl OntologyService {
     /// Validate an RDF file.
     pub fn validate_file(path: &str) -> anyhow::Result<String> {
         match GraphStore::validate_file(path) {
-            Ok(count) => Ok(serde_json::json!({
+            Ok(counts) => Ok(serde_json::json!({
                 "valid": true,
                 "path": path,
-                "triple_count": count,
+                "triple_count": counts.triples,
+                "statement_count": counts.statements,
                 "errors": []
             })
             .to_string()),
--- a/src/shacl.rs
+++ b/src/shacl.rs
@@ -48,6 +48,12 @@ impl ShaclValidator {
             "#,
         )?;
 
+        // sh:targetClass selects SHACL instances, which the specification defines as
+        // reachable by rdf:type followed by zero or more rdfs:subClassOf steps, not by a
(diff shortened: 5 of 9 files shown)


fabio-rovai (changes requested):
Thanks for this — the daemon mode is a genuinely useful feature and the code is competent. Requesting changes because the review turned up several confirmed correctness bugs (each verified against the branch source), most stemming from one design choice: proxying by stringifying parsed clap args back into the line-based batch DSL and re-tokenizing server-side. That lossy round-trip produces bugs 3–5 below, and `/api/batch` already accepts a structured JSON form (`parse_json`) that would eliminate the whole class.

## Confirmed bugs (most severe first)

1. **`src/daemon.rs` (~line 92) — daemon with a config-file token bricks all proxied commands.** `serve-http` falls back to the config token (`resolve_http_token`, main.rs ~1065) and enforces Bearer auth, but `start_daemon` writes `token: null` to daemon.json when `--token`/env isn't passed. Every proxied command then gets 401, and `proxy_batch` hard-fails (`anyhow::bail!`) with no fallback to local execution. One `daemon start` with `[http] token` set in config makes the whole CLI error until `daemon stop`.

2. **`src/daemon.rs` (~line 72) — `--data-dir` split-brain.** `start_daemon` passes `--data-dir` to the child, but the `ServeHttp` arm ignores `cli.data_dir` entirely and derives its store from `cfg.general.data_dir` (main.rs ~1067). With `--data-dir /custom daemon start`, daemon.json lands in `/custom` while the daemon serves `~/.open-ontologies` — proxied and local commands silently see different data. Invisible when both default to the same dir, which is why casual testing wouldn't catch it.

3. **`src/main.rs` (~line 575) — `push` is unusable whenever a daemon runs.** `to_batch_string` serializes it, but `BatchRunner::execute` has no `push` arm — it falls through to `unknown batch command` (batch.rs ~131) and exits 1. It's the only proxied command with no handler.

4. **`src/main.rs` (~line 560) — relative paths resolve against the daemon's cwd, not the caller's.** Paths for `load/save/lint/shacl/ingest/plan/drift` are forwarded verbatim; `start_daemon` never sets `current_dir`, so the daemon inherits wherever `daemon start` ran. `cd /data && onto load ./x.ttl` fails or loads a different file; `save ./out.ttl` writes into the wrong directory. Canonicalize path args client-side before proxying.

5. **`src/main.rs` (~line 616) + `src/batch.rs` (~line 572) — newlines in arguments break proxying.** `parse_lines` splits on `\n` *before* quote-aware parsing, so a multi-line SPARQL query (the norm) is torn across lines → unterminated-quote error via daemon, while the identical command succeeds locally. The double-quote fallback in `shell_quote` also fails to escape backslashes, mangling args containing both quote types.

6. **`tests/cli_test.rs` (untouched by the PR) — 7 existing tests fail on this branch** because the default output changed to human-readable: `test_cli_stats_empty`, `test_cli_validate_file`, `test_cli_drift`, `test_cli_align_feedback`, `test_cli_lint_feedback`, `test_cli_enforce_feedback`, and `test_cli_lint_suppression_end_to_end` (which does `serde_json::from_str(stdout)` and panics). Any external script piping output to `jq` breaks the same way — the default flip deserves an explicit compatibility decision (and updated tests), not a side effect.

7. **`src/output.rs` (~line 84) — successful `daemon start` prints "Daemon dead".** The renderer duck-types on `pid`+`url` presence and defaults `alive` to false; the start-success payload `{ok, pid, url}` matches that branch before the `ok` branch. Root cause: `render_human` guesses the producing command by key-sniffing, and `connect.rs` even strips the batch envelope's `command` field — exactly the discriminator that would make dispatch exact. Suggest keeping the command name through the pipeline and dispatching per command.

8. **`src/main.rs` (~line 1230) — `/api/batch` opens the SQLite StateDb on every request** inside the route closure, while startup already holds a `shared_db` that the adjacent `/lineage` route clones. That's a per-request open on the exact hot path the daemon exists to accelerate.

## Smaller items

- A stray root `package-lock.json` (empty `packages` map, no `package.json` anywhere) is committed — npm accident in a Rust PR; please delete.
- `exec_marketplace` is a third copy of the marketplace list/install logic (main.rs ~1593, server.rs ~702), and the copies already diverge on error shape (server.rs includes available IDs). Extract shared `list`/`install` helpers into `marketplace.rs`.
- The daemon detect-and-proxy block is duplicated between the top of `async_main` and the `Commands::Batch` arm, and the copies already differ (`bail` hardcoded false in one). One helper, one interception point.
- `run_collect` duplicates `BatchRunner::run` line-for-line; `run` can be a thin wrapper over `run_collect`.
- `is_daemon_alive` fork+execs `kill -0` on every proxy-able invocation — a direct `libc::kill(pid, 0)` avoids the spawn. `start_daemon`'s fixed 600ms sleep both wastes time when the port binds fast and reports success when the child failed to bind; polling a TCP connect with a deadline is strictly better.
- No tests for the three new modules (daemon lifecycle, stale-PID cleanup, `render_human` dispatch) — CONTRIBUTING.md asks for tests with new functionality, and this PR has a lot of testable surface.

Happy to re-review once the proxy path moves to the structured JSON form and the lifecycle fixes land.

fabio-rovai:
This sat for five weeks without a word. That is on me, not on you, and there is no good reason for it. Your CI never ran either: workflow runs on fork PRs from first-time contributors park as `action_required` until a maintainer approves them, and nobody did.

I have brought the branch up to date with main. Two conflicts, both trivial and both resolved in your favour by keeping both sides: a CHANGELOG ordering clash, and `src/main.rs` where your `JSON_MODE` static landed in the same place as a `build_main_graph` helper that arrived with the persistent-store work in #81.

Now the part you need, which is what should have reached you in early July.

## The branch does not currently pass

Both of these are on your branch as it stands, independent of my rebase. I checked by testing `717a64f` directly, before merging anything into it, and got the identical result.

**Seven CLI tests fail.**

```
test result: FAILED. 13 passed; 7 failed
```

```
---- test_cli_stats_empty ----
thread panicked at tests/cli_test.rs:63:5:
assertion failed: stdout.contains("classes")

---- test_cli_lint_suppression_end_to_end ----
thread panicked at tests/cli_test.rs:267:68:
called `Result::unwrap()` on an `Err` value: Error("trailing characters", line: 1, column: 3)
```

That second one is the diagnostic. `tests/cli_test.rs` parses stdout as JSON. Making human-readable output the default is a reasonable call, and I am not asking you to reverse it, but it is a breaking change to the CLI contract those tests encode, and the tests were not updated. `cli_test.rs` is not in your changed-files list. The fix is to update those seven to pass `--json` where they parse structured output, and to assert against the new human-readable shape where they are checking for a substring.

**Clippy fails with four errors**, all `collapsible_if` in the daemon-proxy block at `src/main.rs:927-939`:

```
error: this `if` statement can be collapsed
   --> src/main.rs:927:5
    |
927 | /     if !cli.no_connect {
928 | |         if let Some(batch_str) = cli.command.to_batch_string() {
929 | |             if let Some(info) = daemon::read_daemon_info(&cli.data_dir) {
930 | |                 if open_ontologies::daemon::is_daemon_alive(info.pid) {
```

CI runs `cargo clippy -- -D warnings`, so this is a hard gate. Four levels of nesting collapse into one `if` chain with `&&`, or an early return.

## On the feature itself

I have not reviewed the design yet, and I am not going to pretend the above is a design review. What I can say from reading it: the daemon detection being automatic across all 24 graph commands with `--no-connect` as the escape hatch is the right default, and doing it with no new Cargo dependencies is better than I would have expected. The shape looks sound.

Get the tests and clippy green and I will review it properly and promptly. If you would rather not pick this back up after five weeks of silence, say so and I will take the test and lint fixes myself and land it with your authorship intact. Either is fine, and after that delay it is genuinely your call.

For the record, the same audit found #64 and it is now rebased and green on the Rust side.

fabio-rovai:
Apologies for the silence on this one, and thank you for the size of the contribution.

Two mechanical things first. The branch now conflicts with main, which has moved a fair way since 9 August, and CI has never run on it at all: `gh pr checks` reports no checks on `feature/cli-daemon-in-memory-storage`, so nothing here has been through the build, the all-features leg or the Windows leg. A rebase would fix both, but I would rather you did not spend that effort until the two questions below are settled, because one of them could change the shape of the PR.

**Human-readable output by default.** This is the one I want to think about publicly rather than decide in a merge. Anything currently piping `open-ontologies stats` into `jq` breaks the moment this lands, silently and at runtime rather than at a version bump, and CLI output is the one surface where users write scripts we never see. The feature itself is clearly right: the current default is unfriendly for interactive use. The question is only whether the switch is the default now with `--json` to opt back in, or `--human`/`--format text` now and the default flips at 2.0 with a deprecation notice in between. I lean toward the second, but I want to hear the case for the first if you have one.

**Daemon auto-detection on 24 commands.** If a daemon is running, a command silently executes somewhere other than where the user is standing, against a store they may not have loaded. `--no-connect` opts out, which means the safe behaviour is the one you have to know to ask for. I would want the reverse: proxying is explicit, or at minimum every proxied command says on stderr that it ran against the daemon and which store that was. Otherwise the failure mode is a query that answers confidently from the wrong graph, which is exactly the class of defect this repo spends its time removing.

Neither is a rejection. The daemon plus `/api/batch` proxying is a real gap filled, and `exec_marketplace` loading into the daemon store so later commands see the triples is the right instinct. Tell me how you would rather handle the two above and I will review promptly this time.

fabio-rovai:
Decisions on both questions. On the first I have landed somewhere neither of us proposed, and on the second I have changed my mind from what I wrote above, so I will say why rather than quietly reverse.

## Output format: switch on whether stdout is a terminal

I asked you to choose between making text the default now with `--json` to opt back in, and shipping `--human` now with the default flipping at 2.0. Both are worse than the third option, which is what `git`, `ls` and `docker` all do: **human-readable when stdout is a terminal, JSON when it is not.** Add `--json` and `--human` to force either way.

That gives interactive users the readable output immediately, which is the actual problem you set out to fix, and it breaks nothing, because every existing script either pipes or redirects and therefore keeps receiving exactly the JSON it receives today. CI has no TTY, so pipelines are unaffected without anyone editing them. It also makes the deprecation question disappear rather than answering it, and a deprecation window you never have to run is worth more than a well-designed one.

One correction to what I wrote earlier, since I want the reasoning on the record accurately: I said a default flip breaks scripts silently. It does not. `jq` on text input exits non-zero with a parse error, so the break is loud. My objection was to changing a released output contract in a minor version at all, not to the failure being quiet, and TTY detection means we do not have to argue about either.

The one case that needs the escape hatch is a script that runs the CLI attached to a terminal, which is unusual and is exactly what explicit `--json` is for.

## Daemon detection: keep it automatic, make the execution site impossible to miss

I said above that I would want proxying to be explicit, with `--no-connect` inverted. That was wrong, and thinking about your design properly is what changed it.

If proxying were opt-in per command, the feature would be pointless. Nobody wants to pass a flag to all 24 commands. **The user opts in by starting the daemon**, which is already a deliberate act, and after that the session has an execution model. That is coherent and it is what makes `marketplace install` landing in the daemon's store useful at all.

What I still object to is invisibility, and that is fixable without touching the default. A command that runs somewhere other than where the user is standing, against a store they may not have loaded, is capable of answering a query confidently from the wrong graph, and this project exists to remove that class of answer from other people's pipelines rather than add it to ours.

So: **every proxied command announces on stderr** which daemon it reached and which ontology is active there. One line, stderr so no output contract is affected, suppressible with `--quiet` for anyone who finds it noisy. While you are in there, `daemon status` should report what is loaded, not only that the process is alive, since "running" and "holding the graph you think it is" are different facts.

`--no-connect` stays as it is.

## Mechanics

Both of these are additive to what you have written, so the branch does not need reshaping. It does need a rebase, and CI has still never run on it: no checks are reported on `feature/cli-daemon-in-memory-storage`, so nothing here has been through the build, the all-features leg or the Windows leg. Main has moved a long way since 9 August.

Rebase, push, and I will review it promptly. Sorry again for the twelve days.

fabio-rovai:
Both product decisions here were answered on 21 August, so nothing on this PR is waiting on me. Restating them compactly since they are spread across two comments.

Output format goes to TTY detection, which was neither of the options I put to you: human-readable when stdout is a terminal, JSON when it is not, with `--json` and `--human` to force either. That is what git, ls and docker do. Interactive users get readable output immediately, every existing script already pipes or redirects and keeps its JSON, and CI has no TTY so pipelines are untouched. It dissolves the deprecation window rather than choosing between the two lengths of one.

Daemon auto-detection stays, and I reversed my earlier lean to get there. Opt-in per command makes the feature pointless, because nobody passes a flag to twenty-four commands, and the user already opts in deliberately by starting the daemon. The defect was invisibility rather than the default, so the fix is visibility: every proxied command announces on stderr which daemon and which active ontology it reached, suppressible with `--quiet`, and `daemon status` reports what is loaded rather than only that the process is alive. `--no-connect` unchanged.

Both are additive to what you wrote, so there is no reshaping to do.

What is blocking is mechanical. The branch is conflicting against main, which has moved a good deal since 9 August, and CI has never run on it: `gh pr checks` reports no checks at all, so this has not been through the build, the all-features leg, the Windows leg or the Studio typecheck. I have just approved the pending run on the sibling PR to get it a first signal, but there is no run queued here to approve.

Push a rebase and I will approve the run as soon as it appears.

fabio-rovai:
Reopening this is not possible: I force-pushed to the fork branch from the wrong working directory while curating this PR, which pointed `feature/cli-daemon-in-memory-storage` at an unrelated commit. GitHub auto-closed this PR and revoked maintainer push access to the fork, and it refuses to reopen a PR whose head no longer contains its own commits. That was my mistake, not anything wrong with the contribution.

Nothing is lost. The work continues in #120, from a branch in this repository, with your commits preserved in its history: `3ac16f2 CLI: daemon mode, human-readable output, marketplace proxy` is still the parent of everything on top of it, and you are credited as co-author on the follow-up commit.

The eight confirmed bugs from the review are fixed there, the proxy path moved to the structured JSON form as asked, and CI is green. Review of #120 welcome.

PR #78, opened 2026-08-08
fix(serve-http): cancel the shutdown token on ctrl-c and SIGTERM

### LINKED ISSUE #69: serve-http's graceful-shutdown CancellationToken is never cancelled — no SIGTERM/ctrl_c handling
> Context: we're running `serve-http` as a supervised sidecar for Deposium (theseedship); noticed this while testing restart behavior under Docker.

`serve-http`'s axum server is set up for graceful shutdown via a `CancellationToken` (src/main.rs ~958-964, passed into both `StreamableHttpServerConfig` and `axum::serve(...).with_graceful_shutdown(...)`), but nothing ever triggers it — there's no `tokio::signal::ctrl_c()` (or SIGTERM equivalent) anywhere in the binary. In practice this means `docker stop` / `kubectl delete pod` just kills the process outright rather than letting in-flight requests finish or the SQLite state DB flush cleanly.

Would you like a small PR that spawns a task listening for ctrl_c/SIGTERM and calls `ct.cancel()`? The plumbing is already 90% there, it just needs the trigger wired up.

nicolas-geysse:
Closes #69.

`ct` was created at `src/main.rs:958`, cloned into `StreamableHttpServerConfig` at 964, and awaited at 1155 — and never cancelled anywhere. The graceful-shutdown future pended forever, so that path was dead code and `docker stop` killed the process before the state DB flushed. As you put it, that makes it correctness rather than ops polish.

Two commits: `serve-http` as filed, then `serve-unix`, which I raised in the first revision of this PR and have now folded in.

## 1 — serve-http

A task spawned next to the token awaits a termination signal and cancels it. `shutdown_signal()` selects over `ctrl_c()` and, on unix, `SignalKind::terminate()`, with the unix arm behind `#[cfg(unix)]` so `x86_64-pc-windows-msvc` still builds — there the terminate future is `pending()` and only ctrl-c applies. The spawn site carries the comment you asked for.

Verified on Linux, `serve-http --port 18082`, then `kill -TERM`:

```
Open Ontologies MCP server listening on http://127.0.0.1:18082/mcp
Shutdown signal received — stopping HTTP server
```

Exit 0 after 2s.

## 2 — serve-unix

Same gap, plus one of its own. The accept loop in `socket::serve` was unconditionally infinite, so the process could only be killed — and the socket file it bound stayed on disk afterwards. Anything treating the path's existence as liveness then sees a server that is no longer there. The stale-path unlink at bind time papers over it by cleaning up on the *next* start rather than avoiding the leak.

`serve_with_shutdown()` takes a `CancellationToken`, selects it against `accept()`, and unlinks the path on the way out. `serve()` delegates to it with a token nobody holds, so the public signature is unchanged and existing callers — including `tests/socket_test.rs` — keep the accept-forever behaviour.

Verified, `serve-unix --socket /tmp/oo-test.sock`:

| | `SIGTERM` (this PR) | `SIGKILL` (uncatchable — reproduces the pre-fix path) |
|---|---|---|
| process | exits 0 after 2s | killed |
| socket file | **removed** | **left on disk** |

The SIGKILL column is the control: nothing else in the tree unlinks that path, so before this change every stop leaked it.

## What I deliberately did not do, and why

**In-flight connections are still dropped, not drained.** They are detached `tokio::spawn` tasks; draining them properly needs a `JoinSet` plus a deadline, and picking that deadline is a policy decision that belongs to you rather than to this PR. Happy to add it here or separately if you want it.

**No flush claim for serve-unix.** That arm builds no `StateDb` — only an in-memory `GraphStore` — so your point about the state DB not flushing does not apply there. The gains are the clean exit and the unlinked socket, and I did not want to overstate them.

**A failed handler installation logs and pends, rather than `.expect()`.** Your sketch suggested `expect`; a broken registration that resolves immediately would look exactly like a shutdown request and stop the server at startup, which seemed the worse of the two failures. Easy to change back if you prefer the panic.

`cargo clippy --all-targets -- -D warnings` clean. Full suite: 524 passed, 0 failed.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -687,14 +687,14 @@ fn build_tool_filter(
     })
 }
 
-/// Resolves when the process is asked to terminate.
+/// Resolves when the process is asked to terminate, naming the signal.
 ///
 /// Listens for ctrl-c on every platform and, on unix, for `SIGTERM` as well —
 /// the latter is what `docker stop`, systemd and Kubernetes actually send, so
 /// handling only ctrl-c would leave container shutdown unhandled. The unix arm
 /// is behind `#[cfg(unix)]` so the `x86_64-pc-windows-msvc` target in the
 /// release matrix still builds; there the future simply never resolves.
-async fn shutdown_signal() {
+async fn shutdown_signal() -> &'static str {
     let ctrl_c = async {
         if let Err(e) = tokio::signal::ctrl_c().await {
             eprintln!("Failed to install ctrl-c handler: {e}");
@@ -720,8 +720,18 @@ async fn shutdown_signal() {
     let terminate = std::future::pending::<()>();
 
     tokio::select! {
-        _ = ctrl_c => {}
-        _ = terminate => {}
+        _ = ctrl_c => "ctrl-c",
+        _ = terminate => "SIGTERM",
+    }
+}
+
+/// Exit code conventionally reported for a process ended by `sig`.
+fn signal_exit_code(sig: &str) -> i32 {
+    // 128 + signal number: SIGINT = 2, SIGTERM = 15.
+    if sig == "ctrl-c" {
+        130
+    } else {
+        143
     }
 }
 
@@ -1003,9 +1013,20 @@ async fn async_main() -> anyhow::Result<()> {
             tokio::spawn({
                 let ct = ct.clone();
                 async move {
-                    shutdown_signal().await;
-                    eprintln!("Shutdown signal received — stopping HTTP server");
+                    let sig = shutdown_signal().await;
+                    eprintln!("{sig} received — stopping HTTP server");
                     ct.cancel();
+                    // Tokio installs its signal handlers process-wide and never
+                    // removes them, so from here the default terminate
+                    // disposition is gone for good. If graceful shutdown then
+                    // stalls — a long synchronous /api/query or /api/load holds
+                    // its connection open, and axum waits for in-flight
+                    // requests — a second signal would be swallowed and only
+                    // SIGKILL would end the process. Stay listening so the
+                    // second one is decisive.
+                    let sig = shutdown_signal().await;
+                    eprintln!("Second {sig} while shutting down — forcing exit");
+                    std::process::exit(signal_exit_code(sig));
                 }
             });
             // rmcp >=1.4 marks StreamableHttpServerConfig #[non_exhaustive], so it
@@ -1262,10 +1283,20 @@ async fn async_main() -> anyhow::Result<()> {
             let ct = CancellationToken::new();
             tokio::spawn({
                 let ct = ct.clone();
+                let socket_path = socket_path.clone();
                 async move {
-                    shutdown_signal().await;
-                    eprintln!("Shutdown signal received — closing socket");
+                    let sig = shutdown_signal().await;
+                    eprintln!("{sig} received — closing socket");
                     ct.cancel();
+                    // Same reasoning as the ServeHttp arm: tokio's handlers are
+                    // installed for the process lifetime, so a second signal
+                    // would be swallowed if the accept loop were slow to unwind.
+                    // Unlink on the forced path too, so the escape hatch does
+                    // not reintroduce the leaked socket this commit removes.
+                    let sig = shutdown_signal().await;
+                    eprintln!("Second {sig} while shutting down — forcing exit");
+                    let _ = std::fs::remove_file(&socket_path);
+                    std::process::exit(signal_exit_code(sig));
                 }
             });
             open_ontologies::socket::serve_with_shutdown(&socket_path, graph, ct).await?;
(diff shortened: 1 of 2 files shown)


nicolas-geysse:
Folded `serve-unix` in, as flagged at the bottom of the first revision — the PR description is updated to cover both arms.

It had the same missing-cancel gap, and one of its own: the accept loop was unconditionally infinite and the bound socket file was left on disk after every stop, so a liveness check on the path kept seeing a server that was gone. `serve_with_shutdown()` now selects the token against `accept()` and unlinks on the way out; `serve()` delegates with a token nobody holds, so the public signature and `tests/socket_test.rs` are untouched.

Measured rather than assumed — `SIGKILL` is the control, since it cannot be caught and so reproduces the pre-fix path exactly:

| | `SIGTERM` (this PR) | `SIGKILL` |
|---|---|---|
| process | exits 0 after 2s | killed |
| socket file | removed | left on disk |

Three things I left out on purpose, spelled out in the description so they read as choices rather than oversights: in-flight connections are still dropped rather than drained (needs a `JoinSet` and a deadline — that deadline is your call); I make no flush claim for `serve-unix`, since that arm builds no `StateDb`; and a failed handler installation logs and pends instead of `.expect()`, because a registration that resolves immediately would look like a shutdown request and stop the server at startup.

Any of the three is a one-line change in your direction if you'd rather have it the other way.

524 tests passing, clippy `-D warnings` clean.

nicolas-geysse:
Addressed in 1994c5c, with a caveat on the reproduction that is worth stating plainly.

**What I could not reproduce.** I tried twice to stall graceful shutdown and force the second-signal path, and failed both times. An incomplete HTTP request held open: server still exited on the first `SIGTERM`. A real MCP session — full `initialize` handshake, session id, then a held-open SSE stream on `GET /mcp` returning 200: server still exited on the first `SIGTERM`, in about 2s. The reason is in the code: `http_config.cancellation_token = ct.clone()` at `main.rs:964` hands the *same* token to rmcp, so cancelling it tears down the MCP sessions, which lets hyper's graceful shutdown finish. The common cases self-unblock.

**Why I fixed it anyway.** The mechanism the review describes is real regardless: tokio installs its signal handlers process-wide and never removes them, so once that task has run the default terminate disposition is gone. The window that remains is a long synchronous handler — `/api/query`, `/api/load` — holding its connection while axum waits for in-flight requests. And whatever the odds, this PR is what creates the exposure: before it, `SIGTERM` always ended the process immediately. Trading "always dies" for "usually dies, occasionally needs SIGKILL" is not a trade I want to make on your behalf for six lines.

The task now awaits a second signal after cancelling and exits with the conventional 128+signum code. `shutdown_signal()` returns the signal name so the log line and the exit code are both accurate — you get `SIGTERM received — stopping HTTP server` rather than a generic message. In the `ServeUnix` arm the forced path unlinks the socket first, so the escape hatch does not reintroduce the leak the previous commit removes.

First-signal behaviour is unchanged, which is the part worth checking: both transports exit 0 within 2s on a single `SIGTERM`, and `serve-unix` still leaves no socket file behind. 524 tests passing, clippy `-D warnings` clean.

fabio-rovai (approved):
Verified against `e0fc8ec` before merging. The premise holds: `axum::serve(...).with_graceful_shutdown(async move { ct.cancelled_owned().await })` at `src/main.rs:1154-1156` was already wired, and nothing in the tree cancelled the token, so the shutdown future pended forever and the process could only be killed. The plumbing existed and the trigger did not.

Three things I checked specifically, all correct:

- The `#[cfg(unix)]` split on the SIGTERM arm. `x86_64-pc-windows-msvc` is in the release matrix, and `tokio::signal::unix` would not compile there. Falling back to `pending()` is the right shape: the future never resolves, so `select!` degrades to ctrl-c only rather than failing to build.
- Pending on handler-installation failure instead of returning. A broken handler resolving would look identical to a shutdown request and would cancel the token on startup. Easy thing to get backwards.
- The local `use tokio_util::sync::CancellationToken` in the `ServeUnix` arm. The existing import at `src/main.rs:902` is scoped inside the `ServeHttp` arm, so the second one is a separate block scope, not a duplicate.

The second-signal escape hatch is the part I would have asked for if it were not here. Tokio installs signal handlers process-wide and never restores the default disposition, so without it a stalled graceful shutdown would swallow every subsequent SIGTERM and leave SIGKILL as the only exit. Unlinking the socket on that path too is the right call, since the forced path would otherwise reintroduce exactly the leak `serve_with_shutdown` removes.

Keeping `serve` as a wrapper passing a token nobody holds preserves the existing callers without a behaviour change. Good.

Verified locally on the merge of this and #79 together: `cargo test` 524 passed, `cargo clippy --all-targets -- -D warnings` clean, both green in CI once I approved the workflow runs. Note that fork PRs from a first-time contributor sit in `action_required` until a maintainer approves them, which is why these showed no checks for a while. Not your doing.

Thanks for this, and for the commit-message archaeology explaining *why* rather than what.

PR #79, opened 2026-08-08
feat(serve-http): add an unauthenticated /health liveness route

### LINKED ISSUE #68: Add an unauthenticated /health (liveness) route to serve-http, separate from /api/stats
> Context: we're running `serve-http` as a supervised sidecar for Deposium (theseedship) and hit this while wiring container healthchecks.

For anyone running `serve-http` as a supervised sidecar (Docker HEALTHCHECK, Kubernetes readinessProbe), the only close-to-a-health-endpoint today is `/api/stats`, which sits behind the same bearer-token middleware as the real API when a token is configured.

Would you accept a small PR adding a plain `/health` (or `/healthz`) route outside the `/api` nest and outside the auth middleware, returning something minimal like `{"status":"ok","version":...}` — basically what the existing `onto_status` MCP tool already reports, just reachable over bare HTTP? Keeps the probe config decoupled from API credentials.

nicolas-geysse:
Closes #68.

## What this does

`/health` returns `{"status":"ok","version":"<crate version>"}` from `env!("CARGO_PKG_VERSION")` — status and version only, nothing describing loaded state, since an unauthenticated endpoint reporting triple counts is the reason `/api/stats` is the wrong thing to probe. `/health` rather than `/healthz`, and only one of them.

## On the registration point

Your correction about the conditional bearer check was the load-bearing part, and it turns out the ordering has to go one step further than "outer router, not the guarded branch". `Router::layer` wraps only the routes already present, so a route registered *before* the layer call is still wrapped by it even though it lives on the outer router. `/health` therefore has to be registered **after** the `if let Some(ref token)` block, not before it.

I measured both orderings rather than reasoning about them. Server started with `--token s3cr3t-test`:

| | as merged (after the layer) | registered before the layer |
|---|---|---|
| `GET /health` no auth | **200** `{"status":"ok","version":"1.1.1"}` | **401** |
| `GET /health` with auth | 200 | 200 |
| `GET /api/stats` no auth | 401 | 401 |
| `GET /api/stats` with auth | 200 | 200 |

So the wrong ordering reproduces exactly the problem the issue describes, in exactly the deployment where it matters. The code comment records this so the next reader does not have to rediscover it.

## README

There is no `serve-http` section in the README today, so this adds a minimal `<details>` block under "Connect to your MCP client" with a three-row route table and the auth posture of each. Move or reshape it however fits — I did not want to invent a larger section unprompted.

`cargo clippy --all-targets -- -D warnings` clean; `cargo test --lib` 251 passed, 0 failed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,8 +203,16 @@ open-ontologies serve-http --host 127.0.0.1 --port 8080
 | Route | Auth | Purpose |
 |---|---|---|
 | `/mcp` | bearer, if a token is configured | MCP Streamable HTTP — the `onto_*` tools |
-| `/api/*` | bearer, if a token is configured | stats, events and other introspection |
+| `/api/stats`, `/api/lineage` | bearer, if a token is configured | read-only introspection |
+| `/api/query` | bearer, if a token is configured | SPARQL SELECT |
+| `/api/update`, `/api/load`, `/api/load-turtle`, `/api/save` | bearer, if a token is configured | **mutating** — SPARQL UPDATE, and `/load` / `/save` read and write files on the server at a path taken from the request body |
 | `/health` | never | liveness probe |
+
+> **The HTTP surface is not read-only.** When no token is configured the bearer
+> layer is not installed at all, so every route above — including the ones that
+> write to the server's filesystem — is reachable by anyone who can open the
+> port. Set a token, or bind to a loopback address, whenever the port is
+> reachable by anything other than the local process.
 
 `/health` returns `{"status":"ok","version":"<crate version>"}` and is
 deliberately outside the bearer layer, so an orchestrator can probe the process
--- a/README.md
+++ b/README.md
@@ -191,6 +191,28 @@ Add to `.cursor/mcp.json` or equivalent:
 
 </details>
 
+<details>
+<summary><strong>HTTP transport (<code>serve-http</code>)</strong></summary>
+
+For clients that speak MCP over HTTP rather than stdio:
+
+```bash
+open-ontologies serve-http --host 127.0.0.1 --port 8080
+```
+
+| Route | Auth | Purpose |
+|---|---|---|
+| `/mcp` | bearer, if a token is configured | MCP Streamable HTTP — the `onto_*` tools |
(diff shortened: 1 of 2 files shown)


nicolas-geysse:
Good catch from the Codex review, and it was a real problem — fixed in eaaac13.

The route table I added said `/api/*` was "stats, events and other introspection". Checking the router, that is wrong in a way that matters: `/api/update` runs SPARQL UPDATE, and `/api/load` and `/api/save` call `load_file(&path)` / `save_file(&path, &format)` with the path taken from the request body — arbitrary server-side file read and write, not introspection.

Since the point of that table is to help someone decide whether the port needs a token, wording that reads as read-only is worse than no table at all. It is now split by capability, with the mutating routes marked as such, plus a note that when no token is configured the bearer layer is not installed at all — so every one of those routes is open to whoever can reach the port.

Thanks for flagging it; it was my wording, introduced by this PR.

fabio-rovai (approved):
Verified the layer-ordering claim before merging, since the whole security property of this PR rests on it and it is the kind of thing that is easy to assert and easy to get wrong.

It is correct. The bearer middleware is applied via `router.layer(...)` inside the `if let Some(ref token)` at `src/main.rs:1122-1145`. `Router::layer` wraps only the routes registered at the point it is called, so registering `/health` afterwards leaves it outside the auth middleware while `/api` and `/mcp` stay behind it. Registering it before the layer, or inside that branch, would have put it behind the token in precisely the deployment where an unauthenticated probe is the point. The comment explaining this is worth more than the code it sits above; please keep writing those.

Restricting the body to status and version is the right instinct. An unauthenticated endpoint should not describe loaded state, and `/api/stats` remains the authenticated thing to ask for that.

The README table is the part I did not expect and am most glad to have. The warning that the HTTP surface is not read-only, and that `/api/load` and `/api/save` touch the server filesystem at a path taken from the request body, is accurate and was not documented anywhere. That is a real deployment hazard for anyone binding to a non-loopback address without a token, and it is now stated plainly.

Verified locally merged together with #78: `cargo test` 524 passed, `cargo clippy --all-targets -- -D warnings` clean; CI green on both.

Issue #67, opened 2026-08-05
CI never builds/tests the optional feature flags (duckdb, postgres, embeddings, sql, causal-pywhy)


nicolas-geysse:
> Context: we're evaluating open-ontologies as an embedded ontology runtime for Deposium (theseedship), building with `--features duckdb,embeddings` from a pinned commit.

`ci.yml` runs `cargo build` / `cargo test` / `cargo clippy` without any `--features` flag, so `duckdb`, `postgres`, `sql`, `embeddings`, and `causal-pywhy` are never compiled or exercised by CI — a PR could break any of them and CI would still go green.

Would a matrix job (e.g. `cargo build --no-default-features --features duckdb`, `--features embeddings`, `--features sql`) be welcome? For `postgres` you'd presumably need a service container, but `duckdb` and `embeddings` should be self-contained. Happy to draft the workflow addition if useful.

fabio-rovai:
Confirmed. `ci.yml` has exactly three cargo invocations, all bare: `cargo build` (line 31), `cargo test` (33), `cargo clippy -- -D warnings` (35), plus `cargo audit` at 38-39. With `default = []` in `Cargo.toml`, that means `postgres`, `duckdb`, `sql`, `embeddings`, and `causal-pywhy` are compiled by nothing in CI, and the `sqlx`, `duckdb`, `tract-onnx`, `tokenizers`, and `instant-distance` dependency trees are never built at all. A green run currently says nothing about any of it. That is a worse gap than I had assumed and I am glad you raised it.

Worth adding to your list: `release.yml:34` also builds without `--features`, so the published binaries omit duckdb, postgres, and embeddings support entirely. For your use case that is arguably the more immediate problem, since it means the release artifacts cannot do what you are building from source to get. I will decide separately whether the shipped binary should carry the feature set or whether that stays a build-from-source path, but it should not be an accident either way.

Yes please to the workflow addition. To keep it reviewable I would scope the first PR to:

- `cargo check --all-features` on ubuntu as the cheap breadth gate, which covers `causal-pywhy` for free since it has no Rust dependencies.
- `cargo test --features duckdb,embeddings` as the depth gate.
- Leave `postgres` out of this PR. The service container is a fair bit more machinery and I would rather it landed on its own.

[...]

PR #83, opened 2026-08-09
ci(release): attest build provenance and publish SHASUMS.txt

### LINKED ISSUE #77: Add checksums (SHASUMS.txt) or build-provenance attestation to release binaries
Verified against `e0fc8ec`, so after #71 landed.

`release.yml` now populates the release body from the CHANGELOG and uploads the four platform binaries (`release.yml:61-72`). What is still missing is anything that lets a consumer verify the binary they downloaded is the one CI built from that tag: there is no `sha256sum` step, no `SHASUMS.txt` in the assets, and no `actions/attest-build-provenance` or `cosign` step.

GitHub does compute a sha256 digest per asset internally — it is visible through `gh api repos/.../releases/:id` — but that is a platform convenience, not a human-facing artifact, and it is not something a script can pin against ahead of the download.

The concrete case: anyone embedding these binaries into a larger installer or distribution artifact has to take the download on trust, or re-derive a hash and hard-code it with no upstream reference to check it against.

Either of two low-effort additions would close it, both a single step in the existing matrix job:

- `sha256sum open-ontologies-* > SHASUMS.txt` added to the release assets, or
- GitHub's own `actions/attest-build-provenance`, which also ties the artifact to the workflow run and the commit.

Happy to send the workflow patch for whichever you prefer.

(Unrelated to this, but adjacent since it is the same file: `release.yml:34` still builds without `--features`, which is the second half of #67.)

nicolas-geysse:
Closes #77.

Provenance rather than only checksums, as you asked. Both are in, since a `SHASUMS.txt` is still the more convenient thing to pin in a script that does not want a `gh attestation verify` dependency — but the attestation is the one that answers the question the issue was actually about.

## What changes

`actions/attest-build-provenance@v4` signs through Sigstore with the workflow's OIDC identity and binds each artifact to the workflow run and the commit. The job gains `id-token: write` for that identity and `attestations: write` to write the attestation back, on top of the existing `contents: write`.

`SHASUMS.txt` is generated from the renamed binaries and published as a release asset.

Consumers get either of:

```bash
gh attestation verify open-ontologies-x86_64-unknown-linux-gnu --repo fabio-rovai/open-ontologies
sha256sum -c SHASUMS.txt
```

## Three details worth a look

**The attestation runs after the rename.** Before it, all four artifacts are still called `open-ontologies`, so attesting there would sign the same subject name four times and none of them would match a published asset. Placing it after means every attested subject is the name the asset is actually downloaded under.

**SHASUMS.txt is verifiable as-is, not just readable.** The paths in the artifact tree are `open-ontologies-<target>/open-ontologies-<target>`, so a raw `sha256sum` output would list directory-qualified names that no downloader has. The `sed` strips the directory and keeps sha256sum's two-space separator, which is what `-c` parses. Checked against a mock artifact tree including the `.exe`:

```
$ sha256sum -c SHASUMS.txt
open-ontologies-aarch64-apple-darwin: OK
open-ontologies-x86_64-apple-darwin: OK
open-ontologies-x86_64-pc-windows-msvc.exe: OK
open-ontologies-x86_64-unknown-linux-gnu: OK
```

**Pinned to `@v4`**, the current major — worth stating because the widely-copied snippets still say `@v1`.

## Out of scope on purpose

`release.yml:34` still builds without `--features`. That is the other half of #67, and you said you wanted to decide separately whether the shipped binary carries the feature set or whether that stays a build-from-source path. Changing what the published artifact contains is your call, not a side effect of a hardening patch.

CHANGELOG entry under `[Unreleased]`.

I cannot exercise this end to end without pushing a tag on your repo, so the first real run is a tag push here. If the attestation step surprises you, the likely candidate is the subject-path glob against the artifact layout `download-artifact` produces.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -43,6 +43,11 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: write
+      # attest-build-provenance signs through Sigstore with the workflow's OIDC
+      # identity and writes the attestation back to the repository, so it needs
+      # both on top of the existing contents: write.
+      id-token: write
+      attestations: write
     steps:
       - uses: actions/checkout@v4
       - uses: actions/download-artifact@v4
@@ -57,6 +62,19 @@ jobs:
               mv "${dir}open-ontologies" "${dir}open-ontologies-${target}"
             fi
           done
+      - name: Generate SHASUMS.txt
+        run: |
+          # Strip the artifact directory so the listed names match the asset
+          # names as downloaded, and keep sha256sum's two-space separator so
+          # `sha256sum -c SHASUMS.txt` works as-is next to the binaries.
+          sha256sum open-ontologies-*/open-ontologies-* | sed 's|  .*/|  |' > SHASUMS.txt
+          cat SHASUMS.txt
+      - name: Attest build provenance
+        uses: actions/attest-build-provenance@v4
+        with:
+          # Runs after the rename so each attested subject name is the name the
+          # asset is published under.
+          subject-path: 'open-ontologies-*/open-ontologies-*'
       - name: Extract release notes from CHANGELOG
         run: |
           version="${GITHUB_REF_NAME#v}"
@@ -68,4 +86,6 @@ jobs:
         uses: softprops/action-gh-release@v2
         with:
           body_path: release-notes.md
-          files: open-ontologies-*/open-ontologies-*
+          files: |
+            open-ontologies-*/open-ontologies-*
+            SHASUMS.txt
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to Open Ontologies are documented here.
 ## [Unreleased]
 
 ### Added
+- **Build provenance and checksums on release binaries.** The release job now
+  emits a Sigstore attestation via `actions/attest-build-provenance`, binding
+  each published binary to the workflow run and the commit it was built from,
+  and publishes a `SHASUMS.txt` alongside the assets. Verify with
+  `gh attestation verify <binary> --repo fabio-rovai/open-ontologies`, or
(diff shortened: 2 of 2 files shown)


nicolas-geysse:
Legitimate, and it was my instruction that was wrong — fixed in d57b9b6.

`SHASUMS.txt` lists all four platform binaries. Someone who downloaded one of them, which is the normal case, gets three failures next to their single `OK`:

```
$ sha256sum -c SHASUMS.txt
sha256sum: open-ontologies-aarch64-apple-darwin: No such file or directory
open-ontologies-aarch64-apple-darwin: FAILED open or read
sha256sum: open-ontologies-x86_64-apple-darwin: No such file or directory
open-ontologies-x86_64-apple-darwin: FAILED open or read
sha256sum: open-ontologies-x86_64-pc-windows-msvc.exe: No such file or directory
open-ontologies-x86_64-pc-windows-msvc.exe: FAILED open or read
open-ontologies-x86_64-unknown-linux-gnu: OK
sha256sum: WARNING: 3 listed files could not be read
exit 1
```

A valid download reading as a failure is worse than no instruction at all, since the plausible reaction is to assume the binary is bad.

One thing about the suggested remedy, though. `--ignore-missing` is GNU coreutils and is not on macOS by default, and two of the four published binaries are macOS — so making it the advertised command would be the same mistake one platform over. The instruction now selects the single entry, which works with both GNU `sha256sum` and BSD `shasum`:

```
grep open-ontologies-x86_64-unknown-linux-gnu SHASUMS.txt | sha256sum -c -
# macOS: ... | shasum -a 256 -c -
```

`--ignore-missing` is still mentioned, as a GNU convenience rather than as the instruction.

I did not go with per-platform checksum files: four more assets to carry, and the attestation is the better answer for anyone scripting this anyway.

Worth noting this text is what lands on the release page — #71 wires the release body to the CHANGELOG — so the wrong version would have been the first thing a downloader read.

PR #84, opened 2026-08-09
docs(data-pipeline): document the SQL type to XSD contract, versioned

### LINKED ISSUE #76: Document a versioned SQL type → XSD datatype mapping contract for onto_sql_ingest / onto_import_schema
Verified against `e0fc8ec`.

`onto_import_schema` / `onto_sql_ingest` handle both Postgres and DuckDB, including CDC watermarks since #54. The SQL-type → XSD-datatype decision lives entirely in `SchemaMapper::sql_to_xsd` (`src/schema.rs:29-51`) and is not documented anywhere — `docs/data-pipeline.md` documents the *declarative* `datatype` field for CSV/JSON mappings (line 101), which is a different mechanism, and nothing covers the automatic inference.

That matters because the function makes several judgment calls that a downstream integrator would want to know about, and could not discover without reading the source:

- **The catch-all is `xsd:string`** (`schema.rs:49`). Everything unmapped lands there silently — DuckDB `LIST`, `STRUCT`, `MAP`, Postgres `jsonb`, arrays, `interval`, `inet`, enums. A structured column becomes an untyped literal with no warning.
- **Timezone is collapsed.** `timestamp without time zone`, `timestamp with time zone` and `timestamptz` all map to `xsd:dateTime` (`schema.rs:44-45`). XSD does distinguish these, so the offset information is not represented in the range.
- **Precision and scale are dropped.** Parameters are stripped before matching (`schema.rs:33-37`), so `DECIMAL(18,2)` and `DECIMAL(38,10)` both become a bare `xsd:decimal`. Reasonable, but it is a lossy choice worth stating.
- **`uuid` → `xsd:string`** (`schema.rs:48`) rather than `xsd:anyURI` or a custom datatype — again defensible, again invisible.

None of these are bugs. The problem is that they are an implicit contract: a refactor in `schema.rs` would silently change the output shape for every downstream consumer at once, with no changelog entry, because there is nothing that says what the shape was supposed to be.

A short versioned reference — the mapping table, the catch-all behaviour, the timezone and precision decisions, and a note in CHANGELOG whenever it changes — would turn that into an explicit contract at near-zero maintenance cost.

Happy to draft the doc from `schema.rs` as groundwork for a maintainer review, if that is useful.

nicolas-geysse:
Closes #76.

Written against the tree including #80, not against what `main` looked like when the issue was filed.

## Where it lives, and why

`docs/data-pipeline.md`, directly after the declarative `datatype` section, with the distinction you asked for stated up front: one mechanism is written by the operator and honoured as-is, the other is inferred by `sql_to_xsd` and never overrides it. Conflating them is the easy mistake, and having the two sections adjacent is what makes the difference legible.

There is also a pointer from **Schema → ontology in 3 commands**, since that is where a reader arrives when they actually run the import.

## The table is verified exhaustive, not transcribed

I extracted the match arms from `src/schema.rs` and diffed them against the table both ways:

```
arms in code: 39 | rows in doc: 39
missing from doc: none
extra in doc:     none
```

A hand-copied table that drifts one arm is worse than no table, so this seemed worth mechanising rather than eyeballing.

## What it records beyond the table

Each of these is defensible alone and invisible undocumented, which is the argument the issue was making:

- parameters stripped, so `DECIMAL(18,2)` and `DECIMAL(38,10)` are both a bare `xsd:decimal` — with a pointer to SHACL if precision must be enforced;
- IEEE 754 floats are not `xsd:decimal`, with the `NaN`/`INF` reasoning from your fix;
- bare `float` widens to `xsd:double` because the dialects disagree, and widening is the direction that overstates a range without misrepresenting a value;
- timezone is not represented in the range;
- `uuid` is `xsd:string` rather than `xsd:anyURI`;
- the `xsd:string` catch-all, flagged as the row to watch, since DuckDB `LIST` / `STRUCT` / `MAP` and Postgres `jsonb`, arrays, `interval` and enums all land there silently.

It also documents what the import decides **beyond** the datatype, because it changes how the generated ontology reads and none of it is obvious from the type table: foreign key columns never reach `sql_to_xsd` at all (they become an `owl:ObjectProperty` ranged on the parent class), primary keys gain `owl:FunctionalProperty`, and `NOT NULL` adds an `owl:minCardinality 1` restriction.

## Versioning

Marked **v1**, with the rule stated inside the section rather than only in a commit message: a row that changes bumps the version and gets a CHANGELOG entry under `### Changed`. The point being that a `schema.rs` refactor which alters output without one is a breaking change to every downstream integrator simultaneously.

CHANGELOG entry under `[Unreleased]`.

Docs only, no code change. Reshape the placement however suits you — I kept the section self-contained so it can move without dangling.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@ All notable changes to Open Ontologies are documented here.
 ## [Unreleased]
 
 ### Added
+- **Versioned SQL type → XSD datatype contract.** `docs/data-pipeline.md` now
+  documents what `SchemaIntrospector::sql_to_xsd` produces for every SQL type it
+  recognises, alongside the declarative `datatype` mapping field it is easy to
+  confuse it with. The table is marked v1 and any row that changes gets a
+  CHANGELOG entry, so a `schema.rs` refactor can no longer alter the shape of
+  generated ontologies invisibly. Also records the decisions the table encodes
+  (parameters stripped, timezone not represented, `xsd:string` catch-all) and
+  what the schema import derives beyond the datatype.
 - **Opt-in persistent triple store.** A new `[storage]` section selects the
   backend for the main graph: `mode = "memory"` (default, unchanged behaviour)
   or `mode = "persistent"`, which opens a RocksDB-backed Oxigraph store at
--- a/docs/data-pipeline.md
+++ b/docs/data-pipeline.md
@@ -112,6 +112,99 @@ accepted by `onto_ingest` (file rows) and `onto_sql_ingest` (SQL rows).
 When a mapping is omitted, `onto_ingest` / `onto_sql_ingest` auto-generate one
 from the column names so you can iterate fast and refine later.
 
+## SQL type → XSD datatype contract (v1)
+
+The `datatype` field documented just above is **declarative**: you write it, and
+`onto_ingest` / `onto_sql_ingest` honour it. This section is about the other
+mechanism — the **automatic** one. When `onto_import_schema` reads a live
+Postgres or DuckDB schema, nothing declares the datatypes; `SchemaIntrospector::sql_to_xsd`
+(`src/schema.rs`) infers them from the SQL type names. The two are easy to
+conflate and behave differently: a declared `datatype` is never overridden by
+this table, and this table never applies to a mapping you wrote by hand.
+
+Treat what follows as a contract. It is versioned, and any change to a row gets
+a CHANGELOG entry, because a change here silently alters the shape of every
+ontology generated downstream.
+
+### The table
+
+| SQL type (case-insensitive, parameters stripped) | XSD datatype |
+|---|---|
+| `integer` `int` `int4` `bigint` `int8` `smallint` `int2` `tinyint` `int1` `hugeint` `serial` `bigserial` `smallserial` `ubigint` `uinteger` `usmallint` `utinyint` | `xsd:integer` |
+| `numeric` `decimal` | `xsd:decimal` |
+| `real` `float4` | `xsd:float` |
+| `double precision` `double` `float8` `float` | `xsd:double` |
+| `boolean` `bool` | `xsd:boolean` |
+| `date` | `xsd:date` |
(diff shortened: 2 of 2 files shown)


nicolas-geysse:
All three legitimate, all three verified against `src/schema.rs` on `main` — fixed in 1dfe77c.

**The catch-all advice was impossible to follow.** I told the reader to override the inference with an explicit mapping. `generate_turtle(tables, base_iri)` never reads a `MappingConfig` — the string does not occur in `schema.rs` at all — and `import-schema` accepts only a connection and `--base-iri`. Worse, it contradicted the distinction the section opens with: a mapping governs the literals ingest produces later, not the ranges the import asserts. Following my advice would have left the ontology unchanged and the two artefacts disagreeing.

Replaced with the two remedies that do work: cast the column in a view before importing, or amend the `rdfs:range` afterwards with a SPARQL `UPDATE` on the loaded graph, with SHACL for constraints richer than a datatype. I also dropped "untyped literal", which was simply wrong — the import emits `rdfs:range xsd:string`, a typed range.

**The primary-key claim is false for shared primary keys.** The FK branch runs first and emits an `owl:ObjectProperty` only; `is_primary_key` is checked in the `else`. So the 1:1 pattern where a table's PK is also an FK to its parent gets no functional axiom. Now qualified.

**"PascalCase" overstates `table_to_class`,** which splits on `_` alone. Simulated against the real logic:

```
'order_details' → OrderDetails     db:order_details_qty     ok
'order details' → Order details    db:order details_qty     invalid prefixed name
'order-details' → Order-details    db:order-details_qty     invalid prefixed name
```

Documented as a limitation with a sanitisation note.

That last one is worth separating from the doc, though, because it is not really a documentation defect. A quoted identifier containing a space produces `db:order details_qty`, which is not a valid Turtle prefixed name — so `import-schema` against a schema with such a table emits malformed Turtle rather than a clear error. Quoted identifiers with spaces are legal SQL and common in schemas that grew out of spreadsheets.

I have documented it rather than filed it, since you already have five open issues from me and I would rather not add a sixth uninvited. Say the word and I will write it up properly, or send a patch that sanitises the segments in `table_to_class` and the property-name format — though the naming change would be visible in generated output, so it is your call whether that is a fix or a break.

PR #89, opened 2026-08-10
feat(ossie): compile Apache Ossie ontology documents to OWL 2 DL + SHACL

fabio-rovai:
## What

Adds `src/ossie.rs` and the `onto_ossie_import` MCP tool: compiles an [Apache Ossie](https://ossie.apache.org/) (incubating, formerly Open Semantic Interchange) ontology document into OWL 2 DL + SHACL, and optionally loads it so the rest of the crate can work on it.

## Why

Ossie is the vendor-neutral semantic-model spec backed by Snowflake, Salesforce, Databricks, dbt Labs and ~50 other platforms. Its ontology module (`ontology/ontology.md` 0.2.0.dev0) is a fact-based conceptual model: `EntityType`/`ValueType` concepts, relationships with ordered roles, multiplicities, verbalizations, identifiers, derivation rules.

It references neither RDF, OWL, SKOS nor SHACL. A grep for those terms across the whole `apache/ossie` repo returns one incidental hit. So an Ossie ontology is invisible to every reasoner and validator in the semantic web stack.

After `onto_ossie_import` with `load=true`, `onto_reason`, `onto_dl_explain`, `onto_shacl`, `onto_query`, `onto_drift` and `onto_vocab_check` all work on a semantic model that previously had no formal semantics at all.

## Four constructs OWL 2 DL cannot take

These are properties of the logic, not gaps in the implementation, and each is why SHACL is emitted alongside the OWL:

| Construct | Why | Carried by |
|---|---|---|
| `OneToOne` onto a `ValueType` | `InverseFunctionalDataProperty` is forbidden in OWL 2 DL (decidability); OWL Full only | `sh:sparql` |
| `ManyToOne` on arity >= 3 | functional dependency across a tuple; OWL cardinality restrictions are per-property | `sh:sparql` |
| `derived_by` | recursive rule language; OWL has no rule construct | annotation, unenforced |
| `requires` beyond scalar comparison | ungrammared SQL expression string | annotation, unenforced |

The first is the structural one. In fact-based modelling the preferred identifier of an entity type is *by construction* a relationship to a value type, so this is the common case rather than a corner: the reference document declares **three** `OneToOne` relationships and emits **zero** `owl:InverseFunctionalProperty` axioms. Any Ossie-to-OWL path that does not also emit SHACL silently loses every identifier in the model.

Nothing is dropped: everything unenforceable is written to an `ossie:` annotation property in a reserved namespace, all declared `owl:AnnotationProperty` so the graph stays inside OWL 2 DL rather than sliding into OWL Full.

## Verification

24 new tests (17 unit + 7 integration), 268 lib tests green, no regressions.

`tests/ossie_import_test.rs` cross-checks structure counts and the full unenforceable-construct breakdown against an **independent Python implementation** of the same mapping. Both report, on the reference document:

- 14 concepts (6 EntityType, 8 ValueType), 15 relationships
- 16 unenforceable constructs: 4 derivation, 3 inverse-functional-data-property, 3 requires, 2 reified n-ary, 2 n-ary multiplicity, 2 unary-as-class
- 5 SHACL SPARQL constraints
- 10 `owl:Class` in the compiled graph

A divergence in either implementation fails the build rather than producing a quietly different graph. The emitted Turtle is asserted to parse into the oxigraph store and to be SPARQL-queryable.

## Related

The Python side is a converter contributed to `apache/ossie` itself (`converters/rdfowl`), which additionally does the reverse direction and asserts round-trip fidelity through Turtle and JSON-LD.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -68,6 +68,7 @@ Claude dynamically decides the next tool call based on what the previous tool re
 | `onto_ingest` | To parse structured data (CSV, JSON, NDJSON, XML, YAML, XLSX, Parquet) into RDF and load into the store |
 | `onto_sql_ingest` | To run a SQL `SELECT` against PostgreSQL or DuckDB and ingest the result rows into RDF (uses the same mapping format as `onto_ingest`). DuckDB acts as a federation backbone via its `httpfs`/`parquet`/`csv`/`postgres_scanner`/`iceberg` extensions. Connection strings: `postgres://…`, `duckdb:///path.duckdb`, `:memory:`, or `*.duckdb` file path. |
 | `onto_map` | To generate a mapping config from data schema + loaded ontology for review |
+| `onto_ossie_import` | To compile an Apache Ossie (incubating, formerly Open Semantic Interchange) ontology document into OWL 2 DL + SHACL and optionally load it, making a vendor semantic model reasonable/validatable for the first time. Reports the four constructs OWL 2 DL cannot express (`OneToOne` onto a `ValueType`, `ManyToOne` on arity>=3, `derived_by`, non-scalar `requires`) instead of dropping them |
 | `onto_shacl` | To validate loaded data against SHACL shapes (cardinality, datatypes, classes) |
 | `onto_vocab_check` | To closed-world-check generated DATA: flags any predicate/class used that is not declared in the loaded ontology (hallucinated terms). Catches what open-world SHACL silently passes. Run on LLM-generated Turtle before `onto_load` |
 | `onto_reason` | To run RDFS or OWL-RL inference, materializing inferred triples |
--- a/src/inputs.rs
+++ b/src/inputs.rs
@@ -1058,3 +1058,26 @@ mod tests {
         assert!(parsed.struct_dim.is_none());
     }
 }
+
+#[derive(Deserialize, JsonSchema)]
+pub struct OntoOssieImportInput {
+    /// Path to an Apache Ossie ontology document (YAML or JSON) OR the document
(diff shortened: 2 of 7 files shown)


fabio-rovai:
Merging. Recording why, since the title invites the opposite conclusion: the **Apache contribution play was dropped on 10 August** — too much process overhead, and nothing was ever sent (no PR, no mailing list post, no ICLA). This code was kept deliberately. It is an importer that turns Ossie/OSI-style conceptual models into OWL 2 DL + SHACL, and it stands on its own as a product feature whether or not we ever engage with the ASF project.

Worth preserving alongside it, because it generalises well past Ossie: **every identifier in a fact-based / ORM-style semantic model falls outside OWL 2 DL.** A `OneToOne` onto a value type is an `InverseFunctionalDataProperty`, which OWL 2 DL forbids on decidability grounds. In fact-based modelling the preferred identifier of an entity type is by construction a relationship to a value type, so this is the common case rather than a corner — measured here as 3 `OneToOne` declarations producing 0 `owl:InverseFunctionalProperty` axioms. Three other constructs escape OWL the same way: `ManyToOne` at arity >= 3, recursive derivation rules, and constraint languages typed as bare strings.

The consequence is the reason this importer emits both: any conceptual-model-to-OWL path that stops at OWL loses identity semantics silently, so SHACL is mandatory, not optional.

Verified against current `main` rather than this branch's original CI, which predates both the #91 plan/apply work and #90: merged result is 60 suites, 708 tests, `cargo clippy --all-features --all-targets` clean.

Issue #75, opened 2026-08-08
SQLite migrations discard ALTER TABLE errors and there is no schema-version tracking


nicolas-geysse:
Verified against `e0fc8ec`.

Schema evolution past the initial `CREATE TABLE IF NOT EXISTS` set relies on two blocks in `StateDb::open` (`src/state.rs:149-157`):

```rust
let _ = conn.execute_batch(
    "ALTER TABLE monitor_watchers ADD COLUMN webhook_url TEXT;
     ALTER TABLE monitor_watchers ADD COLUMN webhook_headers TEXT;"
);
let _ = conn.execute_batch(
    "ALTER TABLE align_feedback ADD COLUMN signals_json TEXT;"
);
```

Both discard the `Result`. That is clearly deliberate — it swallows the expected "duplicate column name" on an already-migrated DB — but it swallows everything else too: a locked file, a disk error, a partially-applied batch. The first `ALTER` in a batch can succeed and the second fail, and the caller cannot tell.

There is also no schema-version tracking. `PRAGMA user_version` appears nowhere in the tree, so nothing records which migrations have been applied to a given `open-ontologies.db`, and nothing can detect a half-migrated state.

This only bites people who upgrade in place rather than starting from a fresh state DB — which is exactly the long-lived sidecar deployment.

Two things that would close it:

1. A `PRAGMA user_version`-based migration tracker: each migration bumps the version, and the "column already exists" case is handled by the version check rather than by discarding errors, so any *other* error can propagate.
2. A migration matrix test in CI: start from a schema-N-1 DB fixture, run the current binary, assert the new columns exist and prior rows survived. That catches the regression before it reaches someone's real on-disk state.

Would that be a welcome addition?

fabio-rovai:
Confirmed on both halves. `grep -rn user_version src/ .github/` returns nothing, so there is no version tracking anywhere in the tree, and the two `let _ = conn.execute_batch(...)` blocks at `src/state.rs:149-157` are the only migration mechanism there is.

Your reading of the intent is right: those discard the `Result` to swallow the expected duplicate-column error on an already-migrated DB. The cost is the one you identified, and the partial-batch case is the sharpest version of it, because the first `ALTER` in a batch commits and the second does not, and nothing records that the DB is now half-way through a migration.

Both proposals are welcome. Two constraints on the first one:

[...]

PR #90, opened 2026-08-11
feat(embeddings): fingerprint the configuration that produced each vector

### LINKED ISSUE #74: Embedding index cache has no model/provider/revision fingerprint — same-dimension model swaps silently mix incompatible vector spaces
Verified against `e0fc8ec`.

The `embeddings` table (`src/state.rs:99-106`) stores `text_vec`/`struct_vec` alongside `text_dim`/`struct_dim`, and `hnsw_index_cache` (`src/state.rs:113-119`) stores `entries_hash` — a fingerprint over the `(iri, text_vec)` set — plus `entry_count`. Neither table records which embedding provider, model name, or model revision produced the vectors.

The consequence is a silent-corruption path rather than a loud failure. If someone changes `[embeddings] model`, or swaps provider (Ollama → OpenAI, say), and the new model happens to emit the same dimension — 1536 and 768 are both crowded — then `text_dim`/`struct_dim` still match and `entries_hash` is unchanged, because the IRIs and the stored vectors are literally the same bytes. The cache is considered valid. The index is now a mix of two semantically incompatible vector spaces, and the only symptom is that alignment and retrieval quietly get worse.

That is a normal operational event, not an edge case: reconfiguring embeddings after initial setup is something most long-running deployments do at least once. And it is much harder to diagnose than an error, because nothing anywhere says "reindex needed".

Would it be reasonable to add `provider`, `model` and `revision` — or a single composite hash of the three — to both tables, and treat a mismatch as automatic cache invalidation alongside the existing `entries_hash` check? The migration is small, and it converts a silent relevance regression into an explicit rebuild.

Happy to draft it if useful.

nicolas-geysse:
Closes #74. Implements the design from your comment, deviations flagged below.

## The two failure modes

Both are covered, and the second is the one that decided the shape:

1. **The mix** — `persist` DELETEs and rewrites the whole map, so `load_from_db` + `upsert` + `persist` emits a union that looks internally consistent.
2. **Zero new entities** — every stored vector old-model, every *query* vector new-model. The entry set is byte-identical, so `entries_hash` is unchanged and no check over the stored set alone can see it.

(2) is why the fingerprint covers the *configuration* rather than the data, and why strengthening `entries_hash` could never have been the fix. It has its own test: `an_unchanged_entry_set_is_still_invalidated_by_a_config_change`.

## What landed

`model_fp`, a composite hash of `(provider, model, revision)`, on **both** tables, compared **alongside** `entries_hash` and never instead of it. On a mismatch: discard, rebuild, `warn` naming the configuration.

| provider | `model` | `revision` |
|---|---|---|
| `local` | model file name | file identity — size, mtime, sha256 of the head, via `SourceFingerprint` |
| `openai`-compatible | resolved model name | none — the API exposes no revision |

`revision` is an `Option` as you asked. For the local arm its stand-in is the **file contents**, not the path: the download URL and on-disk filename are both stable across a swap done by replacing the file, which is the likely way a local model changes.

`resolve_local_model_path` is extracted out of `TextEmbedderProvider::from_config` so the fingerprint hashes *the file that is actually loaded*. Two copies of that resolution could drift, and a fingerprint over the wrong file is worse than no fingerprint — it would report "unchanged" while the loader picked up a different model.

The fingerprint reflects the **effective** configuration, so `OPEN_ONTOLOGIES_EMBEDDINGS_MODEL` and friends count as changes.

## Deviations, both openai-only, both one line

Say the word and either comes out — neither is load-bearing for the rest.

- **`api_base`** — `provider` is `openai` for *every* OpenAI-compatible gateway, so without it, moving `nomic-embed-text` from a local Ollama to a hosted endpoint keeps the same fingerprint while changing the vector space. That is the issue's own scenario, one layer down. Test: `the_same_model_name_on_a_different_gateway_is_a_different_space`.
- **`dimensions`** — truncates output dimensionality on models that support it. Same model name, different vectors.

## Pre-upgrade rows

`model_fp` is NULL on anything written before the column. **Unknown counts as a mismatch**: treating it as matching would let the corruption survive precisely the upgrade that adds the check. One loud rebuild the first time, then never again. The `warn` says so explicitly rather than leaving it to be inferred.

## Migration

Migration 3 adds both columns; `SCHEMA_VERSION` → 3. Tested against a **hand-written v2 fixture** (`tests/fixtures/state/v2-pre-model-fp.db`) — the DDL was copied from the schema as it stood, not generated from the current `SCHEMA`, per the doctrine in `tests/state_migration_test.rs`. The test asserts the fixture's starting state first, so it fails loudly if the fixture is ever regenerated into something else.

Thank you for taking #75 — the versioned migration path was there waiting and this slotted straight into it.

## Tests

9 unit (fingerprint), 7 integration (guard), 1 migration. Existing `vecstore_test` (15) and `embedding_e2e_test` untouched and green.

Verified non-vacuous: with `fingerprint_matches` reverted to always-`true`, the 4 guard tests fail and the 3 no-regression tests keep passing —

```
vectors_from_another_model_are_not_loaded ................. FAILED
an_unchanged_entry_set_is_still_invalidated_by_a_config_change  FAILED
rows_with_no_recorded_fingerprint_are_rejected_once ....... FAILED
the_poincare_index_cache_is_also_guarded .................. FAILED
vectors_from_the_same_model_are_loaded .................... ok
a_cached_index_is_still_reused_when_the_config_is_unchanged  ok
a_store_without_a_fingerprint_is_unaffected ............... ok
```

That last group matters as much as the first: a guard that invalidates always would pass every test above it while making each restart pay a full rebuild.

`cargo clippy --all-features --all-targets` clean.

## Not covered, deliberately

Swapping `tokenizer.json` alone (same `.onnx`) changes the vectors and goes undetected. Same class of defect, fix is one more `SourceFingerprint` — left out only to keep this to what #74 agreed, not because I think it should stay open. Noted in the module docs and in `docs/embeddings.md`. Happy to add it here or in a follow-up.


nicolas-geysse:
Both legitimate, both fixed in `72195c3`.

## 1. Head prefix → whole file

You are right, and the miss is reachable rather than theoretical: two fine-tunes of one architecture have byte-identical **sizes** *and* byte-identical **ONNX graph protos**, `cp -p` preserves mtime, and the weights that differ can sit past the first 64 KiB. `SourceFingerprint` is the right tool for `ontology_cache` — a missed change there costs a re-parse — and the wrong one here, where it costs the exact corruption the module exists to close. I reached for it because you mentioned it, and did not re-derive whether its cost/benefit transfers. It does not.

The revision is now `sha256` of the whole file. I measured before choosing, on the crate's own implementation, in release:

```
SHA256 over 470 MB: 1.426 s  (330 MB/s)
```

470 MB is the default model (`x-linked-size: 470268510`). Paid once per process, only by the local provider, and against a `TextEmbedder::load` that already reads and optimises that same file — so it is noise next to work already being done.

One wrinkle worth naming: `sha256` copies its input into a padded buffer, so it would have allocated ~940 MB for that model. Added `sha256_reader`, streaming in 1 MiB chunks, sharing the compression function with the one-shot rather than duplicating the arithmetic.

Hand-rolled block padding is exactly the code that is right at 1 MiB and wrong at 64 bytes, so it is tested against the one-shot at every boundary (0, 1, 55, 56, 63, 64, 65, 1 MiB ± 1, 1 MiB + 63/64, …) and against a reader that returns 1, 7, 63, 64, 65 bytes at a time.

`a_same_size_swap_with_preserved_mtime_is_still_detected` reproduces your scenario — same size, identical 64 KiB head, mtime restored — and **fails** when the revision is reverted to `SourceFingerprint`. Checked, not assumed.

## 2. Env tests raced

Correct, and it had not flaked only by luck. `EnvGuard` restored state but granted no mutual exclusion, so `an_env_override_is_reflected` could set `OPEN_ONTOLOGIES_EMBEDDINGS_MODEL` between another test's two `fingerprint` calls.

It now holds a shared `Mutex`, taken **before** the old values are read — I had the ordering wrong in my first pass too: saving them mid-mutation would have restored another test's scratch state as if it were ours.

I kept process-environment mutation rather than injecting the resolvers, because the fingerprint's job is to describe the *effective* configuration and env precedence is part of that. A resolver seam would test the fingerprint against a config struct and leave the precedence path — the thing a deployment actually exercises — uncovered.

---

`cargo clippy --all-features --all-targets -- -D warnings` clean; that flagged a real one on the way (a `MutexGuard` held only for `Drop` reading as a dead field), which would have failed the breadth leg.

fabio-rovai (approved):
Reviewed and verified locally against current `main`, not just the PR's own CI run — `main` has moved a fair bit since you opened this (the #91 plan/apply work touches `state.rs` and `server.rs` too). Merged result: 59 suites, 665 tests, `cargo clippy --all-features --all-targets` clean. Your migration 3 and the new `plans` table coexist without interaction; `state_migration_test` and `vecstore_model_fingerprint_test` both pass on the merge.

On your two open questions: **keep both.** `api_base` is not a deviation so much as the issue's own scenario one layer down — `provider` is `openai` for every OpenAI-compatible gateway, so without it the exact bug #74 describes survives a move from local Ollama to a hosted endpoint. `dimensions` is the same argument. Neither needs justifying by its line count.

Things I checked rather than took on trust:

- The 470 MB hash is paid once per process — `fingerprint()` has a single caller, at server construction — and `sha256_reader` streams in 1 MiB chunks rather than reading the file in. Both of those had to be true for the full-file hash to be the right call, and they are.
- `VecStore::new` has exactly one construction site and it always sets the fingerprint, so there is no path that writes unguarded vectors.
- Non-vacuity: I trust your reverted-guard run, and the three no-regression tests alongside it are the half that usually gets left out. A guard that invalidates unconditionally passes every positive test while making each restart pay a full rebuild — good that it is covered explicitly.

Two small notes, neither blocking:

1. The PR body's table still says `revision` is "size, mtime, sha256 of the head, via `SourceFingerprint`". The code and module docs moved to a full-file sha256 in 72195c3 and the reasoning there is better than the table — worth a description edit so the next reader is not misled by it.
2. `fingerprint: None` skips the checks entirely. That is correct for an unconfigured store and you have documented it, but it means the guard disappears silently if a future caller ever forgets `with_embeddings_fingerprint`. One construction site today, so nothing to do now.

Taking you up on the tokenizer follow-up — filing it separately rather than growing this PR, since you scoped this to what #74 agreed and that was the right instinct. Thanks for the care on the hand-written v2 fixture in particular; generating it from the current `SCHEMA` would have made the test assert nothing.

Issue #95, opened 2026-08-11
Bi-temporal facts: valid-time vs recorded-time on assertions


fabio-rovai:
Roadmap proposal, concept found in Semantica (MIT): every fact carries two independent time dimensions, the interval during which it is TRUE in the world (valid time) and the moment it was RECORDED in the store (transaction time). The registry amendment case makes the value concrete: 'HEK-style cell line X was adherent until 2026-05-01, suspension after' and 'we learned this on 2026-06-12' are different statements, and contradiction detection across documents wants the first while audit wants the second.

Design sketch for an OWL-friendly shape:
- RDF-star or named-graph annotation per assertion: << :X a :AdherentLine >> :validFrom "..." ; :validTo "..." ; :recordedAt "..."
- `onto_query` convenience: AS OF (transaction time) and AT (valid time) query rewriting
- Reasoner: disjointness violations only count as contradictions when valid intervals overlap

Big change (store model + query layer + reasoner semantics), so parked as roadmap. Fits the 4D lineage of the enforce packs (BORO/ies4) unusually well: states with limited temporal extent are exactly what those methodologies model.

nicolas-geysse:
Nice first slice. Named graphs plus validity metadata in the default graph feels like a pragmatic, additive fallback while RDF-star parsing is unavailable, and half-open validity intervals are a good default.

I had a few follow-up thoughts after reading `2479ec5`. None of these need to block the current feature; they could be split into small follow-up issues.

### 1. Assertion shape and round-trip

The first thing I would lock down is a TriG round-trip test. Temporal assertion identity now depends on the named graph, but `GraphStore::serialize` currently iterates over quads and calls `serialize_triple(quad.as_ref())`. That appears to drop the graph name on save/export. Preserving graph names and their temporal metadata feels foundational before adding more semantics.

It would also help to define a few explicit invariants:

- at most one `validFrom`, `validTo`, and `recordedAt` value per assertion graph;
- `validFrom < validTo` when both exist;
- parsed `xsd:date` / `xsd:dateTime` values rather than lexical string comparison;
- a stable assertion identifier and source/authority identity independent of the graph container.

The parsed-time point matters for mixed date/dateTime values, offsets, and invalid literals. Multiple temporal values currently look like last-row-wins during `validities()`, which could make results depend on query iteration order.

### 2. `AT` / `AS OF` query semantics

[...]

PR #105, opened 2026-08-20
test(temporal): conformance corpus for the bi-temporal tools

nicolas-geysse:
Follow-up to #95, first of the additive PRs proposed there. **Tests only, no behaviour change.**

`2479ec5` shipped `onto_temporal_snapshot` / `_query` / `_conflicts` with no temporal tests — `tests/` has no `.trig` or `.nq` fixture. This adds a conformance corpus that pins the *current* semantics so any later change to the temporal layer has to move a line here on purpose:

- **half-open validity** — `[from, to)` with an exclusive upper bound, so the instant a period ends belongs to the next (the correction-not-contradiction case);
- **open bounds** — absent `validFrom`/`validTo` mean "since always" / "still true";
- **`AS OF`** — a fact recorded after the cutoff is not yet visible; a timeless graph carries no recorded time and stays visible;
- **timeless graphs** — a named graph with no validity metadata is always in scope;
- **query scoping** — `query_at` reads only in-scope graphs, and reports an empty scope rather than the whole store;
- **conflicts** — overlapping disjoint assertions are a contradiction; boundary-touching ones are superseded history.

Dates are kept in one lexical form (`YYYY-MM-DD`) on purpose: comparison is currently lexical, and mixed `xsd:date`/`xsd:dateTime`/offset handling is a separate, still-open point on #95.

## Test
`cargo test --test temporal_conformance_test` → 9 passed, 0 failed (default features).

### DIFF
--- a/tests/temporal_conformance_test.rs
+++ b/tests/temporal_conformance_test.rs
@@ -9,9 +9,18 @@
 //! are behavioural anchors, not aspirations: a later change to the temporal
 //! semantics has to move a line here on purpose rather than drift silently.
 //!
-//! Dates are kept in a single lexical form (`YYYY-MM-DD`) on purpose — the
-//! comparison is currently lexical, and mixing `xsd:date` with `xsd:dateTime`
-//! or timezone offsets is a separate, still-open point on #95.
+//! The file has two sections. The first pins semantics that are **intended**;
+//! most of it uses a single lexical date form, since that is the shape those
+//! assertions are about. The second pins what the code does **today** for
+//! mixed-precision and malformed bounds — behaviour that falls out of comparing
+//! datatype-stripped literals lexically rather than being chosen. Each test
+//! there names what the planned parsed-time work (#95, item 4) turns it into,
+//! so that change lands as a diff on assertions that already exist instead of
+//! arriving as new ones the corpus could never have failed on.
+//!
+//! Deliberately not feature-gated: a `#![cfg(feature = ...)]` on a test file
+//! that CI runs without that feature prints `running 0 tests` and reads as
+//! green (this repo lost `tests/schema_test.rs` that way for months).
 
 use open_ontologies::graph::GraphStore;
 use open_ontologies::temporal::Temporal;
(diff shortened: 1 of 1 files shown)


nicolas-geysse:
All three addressed in `081708b` — each was a real coverage gap where the corpus claimed a behaviour it never exercised:

- **open-start intervals**: a described graph with only `validTo` now exercises the `valid_from = None` branch (distinct from the undescribed/timeless one), plus an open-end far-future case;
- **typed literals**: one fixture now uses the `xsd:date` / `xsd:dateTime` literals the tool documents on disk, so the datatype-stripping path in `plain` is exercised (kept single-typed — mixed date/dateTime comparison stays out of scope for this PR, per #95);
- **recorded-time cutoff**: an assertion at `recordedAt` itself now pins the inclusive `recordedAt <= as_of` boundary a strict comparison would break.

12 tests now, all still asserting real output. Thanks Codex.

nicolas-geysse:
Heads-up: the red `features / breadth` and `features / depth` legs here are **not** from this PR — they fail on a pre-existing `clippy::chunks_exact_to_as_chunks` at `src/vecstore.rs:966`, surfaced by a clippy toolchain bump since main's last green run. Fixed in #107 (one line). Once that lands, a rebase turns these green; the rest of the matrix is clean.

fabio-rovai (changes requested):
I want this corpus. #107 is merged, so rebase and the feature legs should go green.

One addition before I merge, per my reply on #95. As written the corpus cannot fail on the thing item 4 changes. The module doc says dates are kept in a single lexical form on purpose, which is defensible for the assertions you are pinning, but it leaves the mixed-precision behaviour untested, so item 4 would land as *new* assertions rather than as a diff on existing ones. That is the opposite of the safety net you described.

Add a second section pinning what the code does today for exactly the cases you plan to change:

- `xsd:date` on one bound against `xsd:dateTime` on the other, same axis
- one bound with a timezone offset, one without
- a bound typed `"01/05/2026"^^xsd:string`, which `plain()` currently strips down to a bare lexical form and sorts against ISO values
- two distinct `validFrom` values on one graph, which `validities()` currently resolves last-row-wins across the UNION

Assert whatever the code answers today, and mark each one in a comment as accidental behaviour that item 4 will flip. Then item 4's PR moves lines that already exist, which is the property that makes the corpus worth having.

Do not feature-gate the file. `tests/schema_test.rs` was `#![cfg(feature = "postgres")]` while CI ran bare `cargo test`, so it printed `running 0 tests` and read as green for months. Yours has no cfg, which is correct, and I would like it to stay that way.

The intended semantics for item 4 are written out in my reply on #95, so you can open it without waiting on me.

nicolas-geysse:
Added in `4e1847d`. You were right that the corpus as written could not fail on the thing item 4 changes — every fixture sat in one lexical form, so item 4 would have arrived as new assertions rather than as a diff.

Second section pins what the code answers **today** for exactly the four cases, measured rather than assumed, each test naming what item 4 turns it into:

| input | today | after item 4 |
|---|---|---|
| `xsd:date` lower vs `xsd:dateTime` upper, same axis | `valid_at=2026-05-01` **in scope** — `"2026-05-01"` is a lexical prefix of `"2026-05-01T00:00:00Z"`, so `instant < to` passes | excluded: the date maps to that exact instant, and the interval is half-open |
| `+02:00` lower bound | `2026-04-30T23:00:00Z` **excluded**, though it is an hour *inside* the interval in real time | in scope |
| `"01/05/2026"^^xsd:string` | **in scope in 1999** — `plain()` drops the datatype and `"0…"` sorts before every ISO value | invalid, with a reason, never timeless |
| two `validFrom` on one graph | one value wins, the other vanishes, nothing reports it as malformed | invalid, both values listed |

One deliberate choice on the multi-valued case: it asserts at an instant *after* both candidate bounds and checks that exactly one value survives, rather than asserting which one. Today the UNION yields `2024-01-01` last, but nothing contracts that order — which is itself the defect — so pinning the winner would have made the corpus flaky rather than strict. The property that flips under item 4 (resolved to a usable interval instead of rejected) is pinned either way.

Module doc rewritten to describe the two sections, and it now records why the file is not feature-gated, with the `tests/schema_test.rs` history as the reason — thanks for that one, it is exactly the failure mode I would have missed.

16 tests, rebased on #107. Clippy `--all-targets` clean locally.

On the rest of your #95 reply: item 3 grows to reporting truncation on all four caps with `validities()` loud rather than a quiet key — agreed, and the distinction you draw is the one that matters, since a truncated `validities()` silently corrupts the scope set rather than shortening a list. Item 4's semantics are clear enough to open without another round trip; I will keep `superseded` and add `non_overlapping` alongside it as you asked, and include `recordedUntil` in the lineage issue. I will open the SHACL-over-a-materialized-snapshot issue ahead of item 5, with the closure question spelled out.

nicolas-geysse:
CI green on all 7 checks after the rebase on #107, including `features / breadth` and `features / depth`. Ready for another look.

fabio-rovai (approved):
This is what I wanted. The four `accidental_*` tests with the ACCIDENTAL / AFTER item 4 comment pairs mean the parsed-time work lands as a diff on assertions that already exist, which is the whole point of a corpus.

`accidental_multi_valued_bound_resolves_to_one_row_instead_of_failing` is better than what I asked for. Pinning "2024-01-01 wins" would have made the test inherit the very iteration-order dependency it is documenting; querying at an instant after both candidate lower bounds and asserting `mentions_early ^ mentions_late` pins the defect without contracting the order. Do that again elsewhere.

Merging. #106 is in too, so item 2 is done and the semantics for item 4 are written out on #95. Item 3 next when you are ready, and please include the `validities()` and `all_graphs()` limits, not just `query_at`.

PR #106, opened 2026-08-20
fix(graph): preserve named graphs when serializing to TriG and N-Quads

nicolas-geysse:
Follow-up to #95. **Additive: dataset formats gain their graph names back; triple formats are unchanged.**

## Problem

`GraphStore::serialize` (`src/graph.rs:296`) rendered every format with `serialize_triple(quad.as_ref())`, which flattens a quad from a named graph into the default graph. Import (`load_file` sniffs `.trig`/`.nq`) and the persistent RocksDB store keep graph names, so the loss was **export-only** — but bi-temporal assertions live in named graphs whose validity is described in the default graph, so a TriG save/reload round trip dropped every `validFrom`/`validTo`/`recordedAt` binding on `onto_save` / `onto_convert` / `onto_pack` output.

## Fix

Branch on the format inside `serialize`:

- **TriG, N-Quads** — `serialize_quad`, which keeps the graph name;
- **Turtle, N-Triples, RDF/XML** — keep `serialize_triple`, which flattens.

The branch matters: a blanket switch to `serialize_quad` would break the triple formats, because `oxrdfio`'s `to_triple` returns an error for a named-graph quad on Turtle/N-Triples/RDF/XML (only default-graph quads convert). Flattening is the correct, only-possible behaviour there, so it stays.

## Test

`tests/graph_named_graph_roundtrip_test.rs`:

- TriG round trip preserves named graphs **and** the default-graph validity metadata;
- N-Quads round trip preserves named graphs;
- N-Triples still flattens (pins the dataset-vs-triple branch, and documents the lossy path);
- a bi-temporal snapshot is still answerable after a TriG round trip — `g1` in scope at 2024-06-01, `g2` not yet valid — which is impossible if the graphs were flattened.

The round-trip tests fail on `main` (flattened export → no named graph to recover) and pass with the fix. CHANGELOG updated under Unreleased → Fixed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,18 @@ All notable changes to Open Ontologies are documented here.
 
 ## [Unreleased]
 
+### Fixed
+- **Named graphs are preserved when exporting to TriG and N-Quads.**
+  `GraphStore::serialize` rendered every format with `serialize_triple`, which
+  flattens a quad from a named graph into the default graph. Import and the
+  persistent store keep graph names, so the loss was export-only — but it meant
+  a TriG save/reload round trip dropped the named-graph structure that
+  bi-temporal assertions live in (issue #95): every `validFrom`/`validTo`
+  binding on `onto_save`/`onto_convert` output was silently gone. The dataset
+  formats now serialize quads; the triple formats (Turtle, N-Triples, RDF/XML)
+  keep flattening, which is the only thing they can represent. TriG and N-Quads
+  round-trip tests in `tests/graph_named_graph_roundtrip_test.rs`.
+
 ### Added
 - **Optional eviction of float32 text vectors from memory**
   (`VecStore::with_text_vectors_evicted`, `turbovec` feature). Without it the
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -296,11 +296,26 @@ impl GraphStore {
     pub fn serialize(&self, format: &str) -> anyhow::Result<String> {
         let store = self.store.lock().unwrap();
         let rdf_format = Self::parse_format(format)?;
+        // Dataset formats carry the graph name; every other format is a single
+        // RDF graph. `serialize_triple` drops the graph name, flattening a quad
+        // from a named graph into the default graph. That is the only thing a
+        // triple format can do, but for the dataset formats it silently
+        // discarded the named-graph structure that temporal assertions live in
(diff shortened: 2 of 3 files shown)


nicolas-geysse:
Heads-up: the red `features / breadth` and `features / depth` legs here are **not** from this PR — they fail on a pre-existing `clippy::chunks_exact_to_as_chunks` at `src/vecstore.rs:966`, surfaced by a clippy toolchain bump since main's last green run. Fixed in #107 (one line). Once that lands, a rebase turns these green; the rest of the matrix is clean.

fabio-rovai (changes requested):
The diagnosis is right and I want this in. #107 is merged, so rebase and the two feature legs should go green.

One change before I merge. Replace the hand-rolled format list in `src/graph.rs` with the API oxrdfio already exposes:

```rust
let carries_graph_name = rdf_format.supports_datasets();
```

oxrdfio 0.2.5 defines it as `pub const fn supports_datasets(self) -> bool` returning `matches!(self, Self::JsonLd { .. } | Self::NQuads | Self::TriG)` (format.rs:133-135). Behaviour is identical today, because `parse_format` (graph.rs:554-565) has no jsonld arm. The difference is what happens when someone adds one: `matches!(rdf_format, RdfFormat::TriG | RdfFormat::NQuads)` would flatten JSON-LD silently, and the round-trip test would not catch it because it only covers trig and nq. Let the upstream crate own the list.

Keep the comment, just retarget the last sentence at `supports_datasets` rather than the enumeration. Everything else is right, including keeping `serialize_triple` for the triple formats, and the round-trip test is the right shape.

nicolas-geysse:
Done in `c9c71af` — `rdf_format.supports_datasets()` replaces the hand-rolled match, and the comment now points at it rather than enumerating formats. Your reasoning is the right one: identical today since `parse_format` has no jsonld arm, but the round-trip test only covers trig and nq, so a hand-rolled list would flatten a later JSON-LD arm silently. Rebased on #107; 4 round-trip + 9 existing `graph_test` cases green locally.

nicolas-geysse:
CI green on all 8 checks after the rebase on #107, including `features / breadth` and `features / depth`. Ready for another look.

fabio-rovai (approved):
`supports_datasets()` is right, and thanks for retargeting the comment at it rather than leaving it describing an enumeration that is no longer there. Merging.

PR #107, opened 2026-08-20
fix(clippy): silence chunks_exact_to_as_chunks in vecstore, keeping the 1.85 MSRV

nicolas-geysse:
The `features / breadth` and `features / depth` CI legs are red on a clippy lint that has nothing to do with the change being reviewed — a clippy toolchain bump promoted `clippy::chunks_exact_to_as_chunks` (stabilized around rust 1.98) to an error under `-D warnings`:

```
error: using `chunks_exact` with a constant chunk size
  --> src/vecstore.rs:966:7
   |
966|     b.chunks_exact(4)
   |       ^^^^^^^^^^^^^^^ help: consider using `as_chunks` instead: `as_chunks::<4>().0.iter()`
   = note: `-D clippy::chunks-exact-to-as-chunks` implied by `-D warnings`
```

`main` was green on these legs on its last run (`37819070`, 2026-08-20) with the older clippy; the fresh toolchain in every new PR run now fails them, so this blocks all PRs (and `main` on re-run), not just one.

The fix is the exact rewrite clippy suggests, semantics-preserving — `chunks_exact(4)` and `as_chunks::<4>().0` both drop the same trailing remainder:

```rust
fn bytes_to_f32_vec(b: &[u8]) -> Vec<f32> {
    b.as_chunks::<4>()
        .0
        .iter()
        .map(|c| f32::from_le_bytes(*c))
        .collect()
}
```

`vecstore` is `#[cfg(feature = "embeddings")]`, so only the feature legs saw it. `cargo check --features embeddings` is clean with this change. Splitting it out as its own one-line PR so it can land ahead of the temporal follow-ups (#105, #106), whose feature legs are red on this and nothing else.

### DIFF
--- a/src/vecstore.rs
+++ b/src/vecstore.rs
@@ -962,6 +962,13 @@ fn f32_slice_to_bytes(v: &[f32]) -> Vec<u8> {
     v.iter().flat_map(|f| f.to_le_bytes()).collect()
 }
 
+// clippy::chunks_exact_to_as_chunks suggests `as_chunks::<4>()`, but that was
+// only stabilized in Rust 1.88, and this crate advertises a Rust 1.85 minimum
+// (README, edition 2024). `chunks_exact` has been stable since 1.31 and decodes
+// identically — it drops the same trailing remainder — so keep it and silence
+// the lint rather than raise the effective compiler requirement for every
+// embeddings build.
+#[allow(clippy::chunks_exact_to_as_chunks)]
 fn bytes_to_f32_vec(b: &[u8]) -> Vec<f32> {
     b.chunks_exact(4)
         .map(|c| f32::from_le_bytes([c[0], c[1], c[2], c[3]]))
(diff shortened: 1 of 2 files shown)


nicolas-geysse:
Good catch — confirmed and fixed. `slice::as_chunks` is `#[stable(since = "1.88.0")]`, and the crate advertises Rust 1.85+ (`README.md:117`, `edition = "2024"`), so the first version would have raised the effective MSRV for every embeddings build. CI runs on `@stable`, so it wouldn't have surfaced it.

Reverted to the original `chunks_exact(4)` decoder — stable since 1.31, identical remainder handling — and silenced the lint with a documented `#[allow(clippy::chunks_exact_to_as_chunks)]` instead, exactly as the lint's own help suggests. MSRV stays at 1.85.

nicolas-geysse:
Second occurrence: the same lint also fires on `tests/turbovec_index_test.rs:531` (an identical byte-to-f32 decode), which only shows up under `--all-targets` with the turbovec feature. Applied the same `#[allow]` there. Grepped the whole tree — those are the only two `chunks_exact` sites. The lib passing clippy on the previous run already confirmed the allow silences it under the CI toolchain; `cargo check --features turbovec --tests` is clean.

PR #113, opened 2026-08-21
fix(cache): cache datasets as N-Quads so named graphs survive a reload

### LINKED ISSUE #112: Compile cache is N-Triples, so a second load of a TriG file silently drops every named graph
Loading a TriG file a second time silently drops every named graph, because the compile cache is written as N-Triples. The first load is correct; every load after it — and every idle-eviction reload — returns a confident, complete, **empty** bi-temporal answer.

This is the import-side sibling of #106. That one fixed `serialize`; this one is the cache round trip on the way back in.

## Measured, against `main@e9e3a99`

One TriG file, three named graphs, validity in the default graph (the shape `src/temporal.rs` documents):

```turtle
ex:g_status_v1 { ex:doc42 ex:status ex:Draft . }
ex:g_status_v2 { ex:doc42 ex:status ex:Published . }
ex:g_owner     { ex:doc42 ex:owner  ex:nico . }
{
  ex:g_status_v1 t:validFrom "2024-01-01" ; t:validTo "2026-05-01" ; t:recordedAt "2024-01-05" .
  ex:g_status_v2 t:validFrom "2026-05-01" ; t:recordedAt "2026-05-02" .
}
```

Two `onto_load` calls on that unchanged file, then `onto_temporal_snapshot` after each, over `serve-http`:

```
LOAD 1  origin = source  →  in_scope = ['g_owner', 'g_status_v1']
LOAD 2  origin = cache   →  in_scope = []
```

After the second load, `SELECT ?g WHERE { GRAPH ?g { ?s ?p ?o } }` returns 0 rows while `SELECT (COUNT(*)) WHERE { ?s ?p ?o }` returns 8: every quad is now in the default graph. `onto_temporal_snapshot` answers `{"ok": true, "complete": true, "in_scope": [], "excluded": []}` — nothing in the response suggests the store lost anything.

The cache file is N-Triples, so the graph names were gone before they were ever read back:

```
$ head -1 ~/.open-ontologies/cache/fixture.d91500914679e0a7.nt
<http://deposium.local/poc#g_status_v2> <https://open-ontologies.org/temporal#recordedAt> "2026-05-02" .
```

## Why

- **Write** — `src/registry.rs:136` (and `:212`, `:367`): `let nt = self.graph.serialize("ntriples")?;`. N-Triples has no way to carry a graph name, so the cache is flattened by construction. With #106 merged, `serialize` correctly branches by format and N-Triples keeps flattening, which is the only thing it can represent — the problem is the format choice here, not the serializer.
- **Read** — `src/registry.rs:126`: `self.graph.load_ntriples(&nt)?` on the cache-hit path. Everything lands in the default graph.
- **Freshness** — `src/registry.rs:106-118`: the key is `(source_path, mtime, size, sha_prefix)`. The file has not changed, so the cache is *correctly* judged fresh. There is nothing wrong with the invalidation logic; the cached artefact is simply not equivalent to the source.

## The path that needs no second load

`ensure_loaded` (`src/registry.rs:226-232`) reloads an idle-evicted ontology from the same `.nt` file:

```rust
if cache_path.exists() {
    let nt = std::fs::read_to_string(&cache_path)?;
    self.graph.clear()?;
    self.graph.load_ntriples(&nt)?;
```

So a long-running `serve-http` with `--idle-ttl-secs` set loses the named graphs by *being idle*, on a single `onto_load`, with no user action at all. The next read tool sees a flattened store.

## Blast radius

Anything whose meaning lives in the graph name. Today that is the whole bi-temporal layer — `onto_temporal_snapshot`, `onto_temporal_query`, `onto_temporal_conflicts` all read `GRAPH ?g` (`src/temporal.rs`), so after a cache-hit load they see a store with no named graphs and report an empty scope as a complete answer. Plain triple ontologies are unaffected, which is why this has stayed invisible.

## Suggested fix

Cache as **N-Quads** rather than N-Triples: line-based and just as fast to parse (the rationale in `src/cache.rs:1-9` survives intact), and it can hold the graph names. `graph.rs` already has both directions — `serialize` accepts `"nquads" | "nq"` (`src/graph.rs:574`) and `load_content(&s, RdfFormat::NQuads)` parses it.

Two details worth deciding rather than discovering:

- **Existing `.nt` cache files.** The extension is built at `src/cache.rs:327`. Moving to `.nq` leaves old entries pointing at `.nt` paths that still exist and still parse — silently keeping the bug for anyone with a warm cache. Either dispatch the reader on the extension, or invalidate on a cache-format version so the first load after upgrade recompiles.
- `src/registry.rs:367` serializes an isolated graph the same way; worth checking whether that path can carry quads too.

Happy to send the PR with a regression test that loads twice and asserts the second load still answers the snapshot — that is the shape that catches it, and no test that loads once ever will.

nicolas-geysse:
Fixes #112.

## What was wrong

The compile cache was written as N-Triples, which cannot carry a graph name, so the cached artefact was not equivalent to the source it stood for. A dataset went in; a flattened graph came out.

```
LOAD 1  origin = source  →  in_scope = ['g_owner', 'g_status_v1']
LOAD 2  origin = cache   →  in_scope = []
```

Two loads of the same unchanged TriG file. After the second, `SELECT ?g WHERE { GRAPH ?g { ?s ?p ?o } }` returns nothing while the triple count is unchanged — every quad moved to the default graph — and `onto_temporal_snapshot` answers `{"ok": true, "in_scope": []}`. The response is not merely empty, it is *confident*: the store genuinely holds no named graphs any more, so nothing downstream has anything to complain about.

The freshness key is not the bug and is unchanged. `(source_path, mtime, size, sha256)` correctly judged the cache fresh; what the cache held was wrong.

**A second path needs no second load and no user action at all.** `ensure_loaded` reloads an idle-evicted ontology from that same flattened file, so a long-running `serve-http --idle-ttl-secs` loses its named graphs by being idle.

## The fix

Cache as N-Quads. That keeps the property N-Triples was chosen for — line-based, fastest to parse, per `src/cache.rs`'s own module doc — and adds the one it lacked.

The extension doubles as the format marker (`CACHE_EXT`), which settles the migration question raised in the issue: an entry pointing at a `.nt` file was written before this fix, holds a flattened dataset, and is **recompiled rather than read**. Dispatching the reader on the extension instead would have kept warm caches silently wrong, which is the one outcome worth avoiding — anyone upgrading has, by definition, a cache full of flattened datasets. The cost is one re-parse per ontology, once.

`GraphStore::load_nquads` is added beside `load_ntriples`, both delegating to one parser body. The existing N-Triples callers (`pack`, `reason`, `tableaux`, `batch`, `main`) are untouched: a single graph is all they carry.

## Tests

Three in `tests/registry_test.rs`, on the existing harness, **each loading twice on purpose**:

- `a_second_load_from_cache_keeps_the_named_graphs` — asserts `origin == "cache"` first, so it fails if the cache silently stops being used rather than passing for the wrong reason;
- `a_reload_after_eviction_keeps_the_named_graphs` — the path that needs no user action;
- `a_cache_file_left_by_an_older_build_is_recompiled_not_read` — writes a legacy `.nt` entry by hand and asserts the next load comes from source.

Graph names are read back through SPARQL rather than from the loader that just claimed to have loaded them.

Mutation-checked rather than assumed: restoring the N-Triples write fails the first two; dropping the format check on the freshness test fails the third. `first_load_writes_cache_file` moves from a `"nt"` literal to `CACHE_EXT` — the extension *is* the format marker now, so the two must not be able to drift apart.

`cargo test` green (70 suites), `cargo clippy --all-targets -- -D warnings` clean, nothing above 1.82 so the 1.85 floor holds.

## Note on scope

The cache-format change also means a `.nt` file left in the cache directory by an older build is now dead weight rather than dangerous. `recompile_named` already removes the previous file when the sha-prefix changes; a sweep for orphaned `.nt` files felt like a separate change and is not here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,34 @@ All notable changes to Open Ontologies are documented here.
 ## [Unreleased]
 
 ### Fixed
+- **The compile cache no longer flattens named graphs** (issue #112). It was
+  written as N-Triples, a format that cannot carry a graph name, so the cached
+  artefact was not equivalent to the source it stood for: a dataset went in and
+  a flattened graph came out. The first load parsed the source and answered
+  correctly; every load after it read the cache back and returned a store with
+  no named graphs at all. Measured on an unchanged TriG file, twice through
+  `onto_load`: `origin: "source"` gave two named graphs, `origin: "cache"` gave
+  none, and `onto_temporal_snapshot` reported `{"ok": true, "in_scope": []}` —
+  a clean, confident, empty answer.
+
+  The freshness key `(source_path, mtime, size, sha256)` was never at fault and
+  is unchanged; the cache was correctly judged fresh, and what it held was
+  wrong. A second path needed no second load at all: an idle-evicted ontology
+  reloads through `ensure_loaded`, from that same flattened file, so a
+  long-running `serve-http --idle-ttl-secs` lost its named graphs by being
+  idle.
+
+  Cache files are now N-Quads (`.nq`) — line-based and just as fast to parse,
+  which was the reason N-Triples was chosen, but able to name a graph. The
+  extension doubles as the format marker: an entry pointing at a `.nt` file was
+  written before this fix, holds a flattened dataset, and is recompiled instead
+  of read. That costs one re-parse per ontology, once, and heals warm caches
+  rather than leaving them quietly wrong. Anything whose meaning lives in the
+  graph name is affected — the bi-temporal tools above all, which read
+  `GRAPH ?g` and saw an empty store. Three tests in `tests/registry_test.rs`,
+  each loading TWICE on purpose: a test that loads once passes on the broken
+  code, which is why this went unnoticed.
+
 - **Named graphs are preserved when exporting to TriG and N-Quads.**
   `GraphStore::serialize` rendered every format with `serialize_triple`, which
   flattens a quad from a named graph into the default graph. Import and the
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -1,9 +1,16 @@
 //! Compile cache for ontology files.
 //!
-//! A loaded ontology is serialized as N-Triples (the simplest, fastest-to-parse
-//! RDF format) and written to `<cache_dir>/<sha>.nt`. Metadata is stored in
-//! the `ontology_cache` SQLite table so we can quickly answer "is the cache
-//! still valid for this source file?" without re-parsing.
+//! A loaded ontology is serialized as N-Quads (line-based and just as fast to
+//! parse as N-Triples, but able to name a graph) and written to
+//! `<cache_dir>/<sha>.nq`. Metadata is stored in the `ontology_cache` SQLite
+//! table so we can quickly answer "is the cache still valid for this source
+//! file?" without re-parsing.
+//!
+//! The format is load-bearing, not a detail. N-Triples cannot carry a graph
+//! name, so caching in it made the cached artefact non-equivalent to the
+//! source: a dataset went in and a flattened graph came out, silently, on
+//! every load after the first. Anything whose meaning lives in the graph name
+//! — the bi-temporal layer above all — lost it there.
 //!
 //! Validity key: `(source_path, mtime_secs, size, sha256(prefix))`.
 //! For most workflows the (mtime, size) pair is sufficient and avoids hashing
@@ -17,6 +24,13 @@ use std::time::UNIX_EPOCH;
 use rusqlite::params;
 
 use crate::state::StateDb;
+
+/// Extension of a cache file, and the format it holds.
+///
+/// N-Quads rather than N-Triples: line-based and just as fast to parse, but it
+/// can name a graph. The extension doubles as the format marker — a `.nt` file
+/// left by an older build holds a flattened dataset and must not be read back.
+pub const CACHE_EXT: &str = "nq";
 
 // Number of head-bytes hashed for the cache fingerprint — overridable via
 // `[cache] hash_prefix_bytes` in config.toml. See `crate::runtime`.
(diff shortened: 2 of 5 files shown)


fabio-rovai:
Verified locally against this branch before merging.

- `cargo test --test registry_test`: 18 pass, including all three new ones, and they execute rather than reporting `running 0 tests`.
- Mutation check on your own claim: restoring `serialize("ntriples")` on the `load_file` write path fails exactly `a_second_load_from_cache_keeps_the_named_graphs` and `a_reload_after_eviction_keeps_the_named_graphs`, and nothing else. The tests are load-bearing.
- `cargo test --all-features`: 71 test binaries, zero failures.

Two things I checked specifically, because the extension guard sits on one path and the reader on another.

`ensure_loaded` calls `load_nquads` with no extension check of its own, and N-Triples is a syntactic subset of N-Quads, so a legacy `.nt` file reaching that reader would parse cleanly and flatten silently: the exact defect this PR closes, surviving on one path. It cannot happen, and it is worth writing down why. The only two sites that ever set `ActiveEntry.cache_path` are the tail of `load_file` and `recompile`, and both derive it from `cache_path_for`, which now always ends in `CACHE_EXT`. A legacy path is therefore recompiled before it can ever be snapshotted, so the guard on the freshness check is sufficient. Nothing to change; I would not have wanted to find that out after merging.

Choosing recompile over dispatch-on-extension is the right call and for the right reason. Anyone upgrading has, by definition, a cache full of flattened datasets, and reading them back correctly would preserve the wrong answer rather than heal it.

One nit I have fixed on main rather than sending you round again: the comment above the reload block still said "Reload from N-Triples cache". In a change whose whole argument is that the cache format is load-bearing, that comment naming the wrong format is the one place a future reader would be misled.

Merged. That is 10 for 10 on claims that verify.

Issue #109, opened 2026-08-20
temporal: a revision-lineage vocabulary — supersedes, retracts, recordedUntil


nicolas-geysse:
Follow-up to #95, on the slice shipped in 2479ec5. This is a design issue: I would like your decision on the shape before writing anything.

## What the vocabulary can express today

`validities()` queries exactly three predicates — `temporal:validFrom`, `temporal:validTo`, `temporal:recordedAt` (`src/temporal.rs:128-132`) — and `Validity` carries exactly those three plus the graph IRI (`src/temporal.rs:53-59`). No assertion identity beyond the graph IRI, no source, no link between one version and the next.

The recorded-time filter is one-sided:

```rust
fn recorded_by(&self, instant: &str) -> bool {
    self.recorded_at.as_deref().is_none_or(|r| r <= instant)
}
```

`src/temporal.rs:69-71`, applied at `src/temporal.rs:180`.

## The observable problem

Two graphs, using only what exists:

```turtle
:g1 { :HEK293 a :AdherentCellLine . }
:g2 { :HEK293 a :SuspensionCellLine . }
{
  :g1 t:validFrom "2024-01-01"^^xsd:date ; t:recordedAt "2024-01-05T00:00:00Z"^^xsd:dateTime .
  :g2 t:validFrom "2024-01-01"^^xsd:date ; t:recordedAt "2026-06-01T00:00:00Z"^^xsd:dateTime .
}
```

`:g2` is a correction: same period, the earlier record was wrong. At `as_of = "2026-07-01"` both pass `recorded_by`, both pass `valid_at("2024-06-01")`, so `snapshot()` returns both in `in_scope` (`src/temporal.rs:179-182`), and `conflicts()` files the pair as a contradiction because the valid intervals overlap (`src/temporal.rs:310-311`).

The general form: `as_of` only narrows forward. It drops graphs recorded *after* the instant; it cannot drop a graph whose belief *ended* before the instant, because no predicate can say a belief ended. So `as_of = now` is the union of every version ever recorded, not the current belief, and "what did we believe then" is answerable only for instants before the first correction. The module doc names that as the question an audit needs (`src/temporal.rs:11-14`).

## Three situations that look identical in the data

- **correction** — one authority replaces its own earlier assertion; the old one should leave the current snapshot and stay reachable at earlier `as_of`.
- **disagreement** — two independent sources assert incompatible things; both must stay in the current snapshot and `conflicts()` should keep reporting them.
[...]

PR #115, opened 2026-08-21
feat(temporal): close the transaction interval with recordedUntil

nicolas-geysse:
First step of #109, on its own, committing to none of the rest of the vocabulary — as you asked.

## What it does

`validities()` reads `temporal:recordedUntil` as a fourth UNION branch, and the recorded predicate becomes two-sided, half-open like `[validFrom, validTo)`.

What was `recorded_by` is now `not_recorded_at`. The second bound gives the predicate a second way to fail, and the two are opposite facts about the assertion: a graph recorded after the audit instant still reads `not yet recorded then`; one whose recorded interval has closed reads `no longer recorded then`. A bool would have flattened them back together at the one place that has to tell them apart, so it returns the reason instead. The old name had exactly one caller and that caller now needs the side, so it folded into the new one rather than sitting beside it unused.

Additive: a store that never writes the predicate is unchanged.

## The two things you asked to be stated

**The cap.** The validity scan counts ROWS, so a fully described graph now costs four of them: 20,000 rows covers **roughly 5,000** such graphs where it covered roughly 6,700 with three predicates (20,000/4 vs 20,000/3 = 6,666.67). A store that does not write `recordedUntil` pays nothing, because nothing here counts graphs. Rather than assert that in a comment, `a_fully_described_graph_costs_four_validity_rows` sets `validity_scan: 3` against a four-predicate graph and requires `complete: false`, then 4 and requires `complete: true`.

**The timeless trap — your catch, and it is the sharp one.** `validities()` builds its map from the predicates it queries, and `scope()` puts anything absent from that map into `in_scope` as `no validity recorded, timeless`, i.e. always true. So a predicate added to the vocabulary and not added to the UNION does not merely go unread — it turns the assertion it describes into an eternal one, which is the worst available default for a bound whose entire job is to withdraw something.

`every_temporal_predicate_takes_its_graph_out_of_the_timeless_bucket` loops over all four predicates, puts each alone in a fixture, and fails if the graph lands in the timeless bucket. It fails today on `recordedUntil`, and it will fail the same way for the fifth predicate someone adds without wiring the UNION.

## Mutation-checked, both directions

| mutation | result |
| --- | --- |
| drop the fourth UNION branch | **4 tests fail**, including the loud one: `` `recordedUntil` does not populate the validity map, so a graph described only by it reads as timeless and always in scope `` |
| `u <= instant` → `u < instant` on the upper bound | **exactly 1 fails**, the half-open test, and nothing else |

## Deliberately not in scope

No `supersedes`, no derivation, no `retracted` bucket, no inverted-interval diagnostic. And the excluded row still carries only `valid` and a `reason` — I did not add the recorded interval beside it, even though an operator reading `no longer recorded then` may want the bound. That is response shape, and this PR is meant to be one predicate and one comparison.

## Verified

`cargo test --lib` 297 pass · `--test temporal_conformance_test` 20 pass (was 16) · `cargo clippy --all-targets -- -D warnings` clean.

Not run through `cargo fmt`: on an untouched `origin/main`, `cargo fmt --check` already reports every file in the repository under rustfmt 1.8.0 here, so running it would bury this diff under a repo-wide reflow. Happy to reformat just these files if your toolchain disagrees with mine.

Docs moved with it: the module TriG example, `src/server.rs` tool description, `CLAUDE.md`, and the `3-way`/`4-way` mention in the corpus.


nicolas-geysse:
Fixed in the latest push — the catch is fair and it lands inside a PR that exists to stop exactly this.

`inputs.rs:1171` and `:1181` are `JsonSchema` doc comments, so they are what a client actually reads out of the generated input schema, and both still promised the cumulative reading: *"Only consider what was recorded by this instant."* With a closing bound that is no longer what happens — an assertion recorded before the cutoff and retired by it is now excluded, so a caller following the published contract would have silently lost withdrawn historical records.

Both now say what the code does: recorded by then **and not retired by then**. I had updated the tool description in `server.rs` and the table in `CLAUDE.md` and missed the two field descriptions, which are the ones that reach clients.

Same class of defect as the false `note` in #116 — output that nobody pinned, drifting from the code. Worth naming plainly.

One thing I did **not** add: a test pinning those schema descriptions. There is no schema-assertion harness in `tests/` today, and inventing one felt like scope creep on a PR meant to be one predicate and one comparison. If you would rather it existed, it is a natural follow-up and I am happy to write it.

Issue #110, opened 2026-08-20
temporal: add a `non_overlapping` bucket to `onto_temporal_conflicts`, deprecate `superseded`, drop it at 2.0


nicolas-geysse:
Follow-up to #95, on the slice shipped in 2479ec5. Small and mechanical; opening it separately so the deprecation window is on the record.

## What the code proves

`conflicts()` collects candidate pairs with one query (`src/temporal.rs:271-280`): same subject typed in two different named graphs, `FILTER(STR(?a) < STR(?b))`, with an `owl:disjointWith` between ancestors reached by `rdfs:subClassOf*`. Same-graph pairs are dropped at `src/temporal.rs:291-293`.

Each surviving pair is then sorted by exactly one test (`src/temporal.rs:310-314`):

```rust
if va.overlaps(vb) { conflicts.push(entry); } else { superseded.push(entry); }
```

`Validity::overlaps` (`src/temporal.rs:75-82`) is interval intersection on the half-open bounds, with an absent bound treated as open. It proves one thing: the two periods share no instant. It does not test adjacency (`valid_to == other.valid_from` is never compared), and `recorded_at` is not consulted anywhere in `conflicts()` — `recorded_by` has its only call site in `snapshot()`, at `src/temporal.rs:180`.

So `superseded` (`src/temporal.rs:321-322`) means "these two disjoint types are never asserted at the same instant". Three different situations land there identically: a genuine correction; two independent sources describing adjacent-but-different periods; and two periods with a gap between them, i.e. missing coverage. The `note` at `src/temporal.rs:323-325` states the narrower claim — "one period ends where the other begins" — which the code does not check. The tool description says the same thing ("everything else as superseded history", `src/server.rs:1033`), as does `CLAUDE.md:100`.

## Why it matters

`superseded` asserts lineage. The data carries none. And once lineage exists (separate issue on #95), a genuinely superseded pair becomes distinguishable from a merely non-overlapping one, at which point the key is not just loose but wrong.

## Proposed shape (additive, no rename)

Per your call on #95 — a rename would break anyone reading 1.2.0 output:

- add `non_overlapping` and `non_overlapping_count`, carrying exactly the values `superseded`/`superseded_count` carry today;
- keep `superseded`/`superseded_count` emitting the same values, marked deprecated in the tool description (`src/server.rs:1033`) and in the CHANGELOG, where the released contract is described (`CHANGELOG.md:165-171`);
- drop both at 2.0.

[...]

PR #116, opened 2026-08-21
fix(temporal): report non_overlapping instead of claiming supersession

nicolas-geysse:
Issue #110, in the additive shape you approved.

## The three decisions, as taken

**1. No opt-out flag.** `superseded` and `superseded_count` are emitted unconditionally until 2.0, carrying exactly the same rows as the new key. `the_deprecated_key_carries_the_same_set_until_2_0` asserts that across all three conflict fixtures, so no reader has to change anything on the day the new key appears.

**2. `superseded` stays gone rather than being redefined.** The comment in the code says so explicitly, so that whoever implements lineage-backed supersession finds the constraint at the point of temptation: it takes a new key, because a key that names a different set on either side of a major version silently changes the answer for a client reading across the upgrade.

**3. The note is fixed now.** It was not a naming preference, it was false output — it claimed "one period ends where the other begins", which is adjacency, when the only test performed is `!overlaps`, which is disjointness. `src/server.rs:1033` and `CLAUDE.md:100` repeated the same claim and move with it.

## The scope addition

The bucket holds three situations and only one is a correction: touching, never-met, and **separated by a gap**. The gap is the one that is not benign — nothing is asserted about the interval at all, so reading it as history invents a continuity nobody wrote down.

It gets no bucket of its own here, as you said. It is named in the tool description and in the note, which is the difference between a caller knowing and a caller finding out.

## The test that should have existed

Nothing pinned the note's text. That is how a false sentence survived a release: it was output, and output nobody asserts on drifts freely. `the_note_claims_only_what_the_code_checks` now pins what the code proves — and restoring the old wording fails it, **and only it**:

```
test the_note_claims_only_what_the_code_checks ... FAILED
  the note must state the check that was performed: … superseded pairs are
  corrections: one period ends where the other begins …
test result: FAILED. 18 passed; 1 failed
```

`CONFLICT_GAP` pins the case the old name got wrong: `contradiction_count == 0`, `non_overlapping_count == 1`, and the same pair lands in exactly the bucket a touching pair does — which is the argument for the rename in executable form.

## Verified

`cargo test --lib` 296 pass · `--test temporal_conformance_test` 19 pass (was 16) · `cargo clippy --all-targets -- -D warnings` clean.

Not run through `cargo fmt`, same as #115: on an untouched `origin/main`, `cargo fmt --check` already reports every file in the repository under rustfmt 1.8.0 here.

## Ordering

Independent of #115 — that one touches `validities()` and the `timeless` literal, this one touches `conflicts()` and its output. They do not collide. #115 first only because you asked for the `recordedUntil` step on its own.


nicolas-geysse:
Fixed in the latest push, and the catch is the sharp one — it is the same defect this PR is about, one level down.

The new key promised its pairs *"are never asserted at the same instant"*. They are not. `overlaps()` compares datatype-stripped literals as **text**, so bounds written with different timezone offsets are ordered lexically rather than on the timeline:

```
ex:g_a  validTo   "2026-05-01T00:00:00-02:00"   =  02:00Z
ex:g_b  validFrom "2026-05-01T01:00:00Z"        =  01:00Z   → they share an hour
lexically: "…T01:00:00Z" > "…T00:00:00-02:00"   → read as disjoint
```

The pair lands in `non_overlapping`, which is the one thing that bucket is supposed to exclude.

**I did not normalise here.** That is item 4 of #95 — already written against `validities()` and the `Validity` comparison — and doing it in this PR would collide with it. The fix is to stop promising more than `overlaps()` can establish, which is precisely the discipline that made `superseded` unusable: a name, or a note, asserting more than the check performed. Renaming the key and then leaving an over-strong guarantee in the note would have reproduced the defect under a better word.

So the note, the tool description and `CLAUDE.md` now qualify the guarantee by the comparison that backs it, and name the offset case explicitly.

And `accidental_offset_bounds_can_hide_a_real_overlap_from_the_disjointness_check` pins today's classification, in the style you asked for on #105: item 4 flips it to `contradiction_count: 1` / `non_overlapping_count: 0` as a **diff on an assertion that already exists**, rather than arriving as a new one the corpus could never have failed on.

`cargo test --lib` 296 · `--test temporal_conformance_test` 20 (was 16) · clippy clean.

fabio-rovai:
Merged as 625c9e5 on main, with your two commits squashed and your authorship on the commit.

Landed by hand rather than through the button because #114 and #115 went in first and left a CHANGELOG conflict, two adjacent `### Fixed` entries, nothing else. I kept both, `recordedUntil` above `non_overlapping`. No code hunk conflicted.

Verified on the rebase rather than on your branch, since the rebase is what shipped: `--lib temporal` 11 pass, `temporal_conformance_test` 24 pass, `clippy --all-targets -D warnings` clean.

I mutation-checked the note test, because it is the one carrying the argument. Restoring the old sentence verbatim into `conflicts()` fails `the_note_claims_only_what_the_code_checks` and only it: 23 passed, 1 failed. A prose `note` was never a contract, but it was output, and output nobody asserts on drifts; that test is the thing that stops it drifting back.

The scope addition is the part I think earns the most beyond #110. Touching, never-met and gap-separated are three different situations, and the gap is the one that is not benign, because nothing is asserted about the interval at all, so reading it as history invents a continuity nobody wrote down. `CONFLICT_GAP` puts that in executable form: same bucket as a touching pair, which is the rename argument without any prose.

`accidental_offset_bounds_can_hide_a_real_overlap_from_the_disjointness_check` is doing exactly what section 2 was built for. #117 flips it to a contradiction as a diff on a line that already exists.

Issue #108, opened 2026-08-20
temporal: scope SHACL validation and OWL inference to a materialized snapshot


nicolas-geysse:
Follow-up to #95, on the "validate/infer over a snapshot" item. Opening this ahead of the snapshot manifest, as agreed.

## What happens today

The two entry points select data in opposite, equally wrong ways, and neither knows about temporal scope.

**`onto_reason` unions every version.** `Reasoner::run` reads the store through `graph.all_triples()` (`src/reason.rs:87`), which iterates `store.iter()` over every quad and keeps only subject/predicate/object — the graph name is dropped (`src/graph.rs:490-501`). Every temporal version of an entity lands in one closure. Worse, materialisation writes the result back with `graph.load_ntriples(&ntriples)` (`src/reason.rs:449`), parsed as N-Triples so the triples go to the **default graph** (`src/graph.rs:406-415`). That is the graph the temporal layer reads validity from (`src/temporal.rs:128-132`), and a graph with no validity description is timeless and in scope for every snapshot (`src/temporal.rs:177`). So a closure computed over a state that never existed is written into the default graph — the one graph `snapshot()` never lists, since `all_graphs()` enumerates named graphs only (`src/temporal.rs:158-163`). The asymmetry that follows is the sharp end of this issue: those inferences are invisible to `onto_temporal_query`, which reads only in-scope named graphs, and permanently visible to `onto_shacl`, which reads only the default graph.

**`onto_shacl` sees no versions at all.** Constraints are emitted as plain patterns — `?focus a <targetClass>` with no `GRAPH` block (`src/shacl.rs:149-153`, `src/shacl.rs:736-739`) — and run through `graph.sparql_select` (`src/shacl.rs:674-678`), which leaves the evaluator on its default dataset specification (`src/graph.rs:199-204`). That specification is the store's default graph only (spareval 0.2.6, `QueryDatasetSpecification::new`), and nothing under `src/` calls `dataset_mut()` or `set_default_graph_as_union()`. Assertions held in named graphs are invisible to the validator.

[...]

Issue #118, opened 2026-08-22
temporal: an empty or inverted validity period overlaps everything, while holding at no instant


nicolas-geysse:
`Period::overlaps` and `Period::valid_at` disagree about a period that holds at no instant, and they have disagreed since before the bi-temporal work. Codex found it reviewing #117; measuring it showed the defect is older than that PR, so it belongs here rather than in it.

## The disagreement

```rust
// src/temporal.rs
fn valid_at(&self, instant: DateTime<Utc>) -> bool {
    self.from.is_none_or(|f| f <= instant) && self.to.is_none_or(|t| instant < t)
}

fn overlaps(&self, other: &Period) -> bool {
    let start_before_end = |a, b| match (a, b) {
        (Some(start), Some(end)) => start < end,
        _ => true, // an open end never closes the interval
    };
    start_before_end(self.from, other.to) && start_before_end(other.from, self.to)
}
```

For `P = [t, t)` — half-open, so true at no instant — `valid_at` is false everywhere, while `overlaps` is true against any period containing `t`, and **unconditionally true** against a timeless one, because both arms fall through to `_ => true`.

An inverted period `[t₂, t₁)` with `t₂ > t₁` behaves the same way. There is no `from < to` check anywhere in the file.

## What it does to the tools

`onto_temporal_snapshot` and `onto_temporal_conflicts` answer differently about the same graph. Measured on `main`:

```trig
ex:g_never  { ex:X a ex:Adherent . }
ex:g_around { ex:X a ex:Suspension . }
{
  ex:Adherent owl:disjointWith ex:Suspension .
  ex:g_never  t:validFrom "2024-01-01" ; t:validTo "2024-01-01" .
  ex:g_around t:validFrom "2023-01-01" ; t:validTo "2025-01-01" .
}
```

`onto_temporal_conflicts`:

```json
{ "contradiction_count": 1,
  "contradictions": [ { "graphs": ["…/g_never", "…/g_around"],
      "periods": ["2024-01-01 to 2024-01-01", "2023-01-01 to 2025-01-01"],
      "subject": "X", "types": ["Adherent", "Suspension"] } ] }
```

`onto_temporal_snapshot` at every instant tried, including `2024-01-01` itself:

```json
"excluded": [ { "graph": "…/g_never", "reason": "not true at that instant",
                "valid": "2024-01-01 to 2024-01-01" } ]
```

[...]

PR #121, opened 2026-08-26
fix(shacl): node-shape constraints reach the tri-state verdict too

nicolas-geysse:
Fixes the `sh:closed` half of #108, ahead of the snapshot work as asked.

## What was wrong

The check added in 5e5fc91 routes unimplemented constraints to `skipped_constraints` and suppresses the verdict, but it reached one `sh:property` hop below the shape and no further: shape discovery projects only the shape and its target class, and both per-shape queries go through `sh:property`, so a constraint asserted on the `sh:NodeShape` subject itself never bound the predicate the complement inspects. Measured on main:

| shape carries | `conforms` | `skipped_constraints` |
| --- | --- | --- |
| `sh:not` on a property shape | `null` | 1 |
| `sh:not` on the node shape | **`true`** | 0 |
| `sh:closed true` on the node shape | **`true`** | 0 |
| `sh:nodeKind sh:IRI` on the node shape | **`true`** | 0 |

## What this does

A second complement query covers the shape node (`src/shacl.rs`). The whitelist is what the validator reads there, the four target forms (the three unimplemented ones are already recorded as skipped at the target check and are not counted twice), `sh:property` and `sh:sparql`, plus the annotation predicates (`sh:message`, `sh:severity`, `sh:name`, `sh:description`, `sh:order`, `sh:group`), which are never constraints. Anything else in the `sh:` namespace lands in `skipped_constraints` with its own reason and forces `conforms: null`, including a construct a later change starts to evaluate, since a whitelist that tracks what is implemented drifts the first time someone adds a constraint and forgets the list. `sh:deactivated` is deliberately among the skipped: SHACL says a deactivated shape must not be evaluated, this validator evaluates it, so the predicate is not honoured and must not read as if it were.

Two decisions worth your eye, both reversible in a line:

1. **The complement is restricted to the `sh:` namespace.** An implicit class target (a node that is both `sh:NodeShape` and `rdfs:Class` / `owl:Class`, accepted since 5e5fc91) carries the class's own axioms on the same subject, `rdfs:subClassOf`, `rdfs:label`, `owl:equivalentClass`; an unrestricted complement reported those as constraints and turned every such run undetermined (measured before adding the guard). A constraint is by definition a predicate in the `sh:` namespace; a predicate from any other namespace on a shape node is an annotation or an axiom. The cost: a non-`sh:` predicate on a node shape is never reported as skipped.
2. **The shape is bound as a variable and matched to the discovered shapes by its printed term, not spliced into the query text.** A shape written `[] a sh:NodeShape` prints as `_:label`, and a blank-node label inside a SPARQL query is a non-distinguished variable, not a name: the first cut of this PR spliced it, enumerated every predicate in the shapes graph, and turned a verdict that was a boolean on main into null for every anonymous shape. It runs once over `SELECT DISTINCT ?shape ?pred`, so a shape with two `sh:targetClass` values records its `sh:closed` once rather than once per target class. It stays restricted to discovered shapes, like the property complement: a shape with no target selects no focus nodes under SHACL, so its constraints not running is what the specification asks for.

The generalisation you asked for is written next to the verdict rather than in a commit message: a validator has three answers, not two, and every path that can select nothing or execute nothing has to reach the third one; `nothing_matched` for a target that selects nothing, `skipped` for a target form that is not implemented and for a constraint that cannot execute, at property level and at node level.

## Tests

`tests/shacl_test.rs`, seven new, all mutation-checked (each named mutation reddens exactly the tests that carry its claim, then the tree is restored):

- `test_tri_state_verdict_is_reachable_from_node_shape_property_shape_and_target`: the null verdict is reachable from the node shape (`sh:closed`), from a property shape (`sh:not`) and from a target (`sh:targetNode`), each with its own `skipped` entry naming the construct. Adding `sh:not` to the property whitelist or deleting the `sh:targetNode` target check reddens it.
- `test_node_shape_constraints_reach_the_tri_state_verdict`: `sh:closed`, node-level `sh:not` and `sh:nodeKind` each land in `skipped` under their own IRI. Disabling the node complement reddens it.
- `test_node_shape_annotations_and_class_axioms_keep_a_boolean_verdict`: annotations on an explicit shape and class axioms on an implicit class target keep a boolean verdict and no `skipped_constraints` key. Dropping the namespace guard reddens the implicit-class case; dropping `sh:name` from the whitelist reddens the explicit one.
- `test_deactivated_shape_is_recorded_as_skipped`: whitelisting `sh:deactivated` reddens it.
- `test_blank_node_shape_is_bound_as_itself_not_as_a_wildcard`, `test_node_constraint_is_recorded_once_however_many_target_classes`, `test_node_constraint_on_a_shape_without_target_is_not_a_gap`: the three consequences of decision 2.

Full default suite in a clean target directory: 74 binaries, 0 failures (`shacl_test` 19). `clippy --all-targets -D warnings` clean apart from `collapsible_else_if` at `src/main.rs:867`, which clippy 1.93 raises on main itself since fccfeb4; left alone. `rustfmt --check` reports the same pre-existing hunks on `src/shacl.rs` and `tests/shacl_test.rs` as on main, none in the added lines, and I did not reformat the files.

Not touched, on purpose: `check_shapes`, which is the other half of #108 and its own PR; and the inherited duplication of the property complement per discovery row, which predates this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,28 @@ All notable changes to Open Ontologies are documented here.
 ## [Unreleased]
 
 ### Fixed
+- **A constraint asserted on the node shape itself was never evaluated and
+  never reported, so `sh:closed true` over data carrying an undeclared
+  predicate returned `conforms: true`.** The check that routes unimplemented
+  constraints to `skipped_constraints` reached one `sh:property` hop below the
+  shape and no further: `sh:closed`, a node-level `sh:not`, `sh:nodeKind`,
+  `sh:and`, `sh:or`, `sh:xone`, `sh:in` and `sh:node` never bound the
+  predicate it inspects. A second complement now covers the shape node, with a
+  whitelist of the predicates the validator reads there (the target forms,
+  `sh:property`, `sh:sparql`) plus the annotation predicates, which are never
+  constraints, so any other `sh:` predicate lands in `skipped_constraints` and
+  the verdict becomes null. The shape is bound as a query variable and matched
+  to the discovered shape, not spliced into the query text: a shape written
+  `[] a sh:NodeShape` is a blank node, and a blank-node label inside a SPARQL
+  query is a wildcard, not a name. The complement runs once per shape rather
+  than once per target class, so a node constraint is recorded once.
+  `sh:deactivated` is among them on purpose: a deactivated shape is still
+  evaluated here, so the predicate is not honoured and must not read as if it
+  were. The complement is restricted to the `sh:` namespace, because an
+  implicit class target carries its own class axioms on the same subject and
+  those are not constraints. A test now asserts the null verdict is reachable
+  from the node shape, from a property shape and from a target, each naming
+  its construct, so the next construct added has somewhere obvious to fail.
 - **A daemon started with `[http] token` in config rejected every command it was
   meant to serve.** `serve-http` falls back to the config token and then enforces
   bearer auth, but `daemon start` recorded `token: null` in `daemon.json`
--- a/src/shacl.rs
+++ b/src/shacl.rs
@@ -2,7 +2,7 @@ use crate::graph::GraphStore;
 use oxigraph::io::{RdfFormat, RdfParser};
 use oxigraph::sparql::{QueryResults, SparqlEvaluator};
 use oxigraph::store::Store;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::io::Cursor;
 use std::sync::Arc;
 
@@ -15,8 +15,11 @@ use std::sync::Arc;
 ///
 /// Any constraint the validator cannot execute is recorded in
 /// `skipped_constraints` and suppresses the conformance verdict: `conforms`
-/// becomes null rather than true. Reporting success for rules that were never
-/// run is the one failure mode this validator must not have.
+/// becomes null rather than true. That holds wherever the constraint sits:
+/// on a property shape (`sh:not`), on the node shape itself (`sh:closed`,
+/// `sh:nodeKind`, `sh:deactivated`) or in a target form that is not selected
+/// (`sh:targetNode`). Reporting success for rules that were never run is the
+/// one failure mode this validator must not have.
 pub struct ShaclValidator;
 
 impl ShaclValidator {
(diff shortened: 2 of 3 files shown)


fabio-rovai:
Codex is right on the P2, and I checked it against the diff rather than taking the bot's word for it.

The complement query binds `?o` and never reads it, so the filter is on the predicate name alone. That is correct for `sh:closed true`, for a node-level `sh:not`, for `sh:nodeKind` and for every other construct on your list. It is wrong for exactly two inputs, and both of them are inputs a real shapes graph carries:

- `sh:closed false` is the SHACL default. It imposes no closed-shape restriction, so there is nothing to evaluate and nothing was missed.
- `sh:deactivated false` says evaluate this shape, which is what the validator does. The predicate is honoured in full.

In both cases the run evaluated everything the shape asked for, and the report comes back `conforms: null` with a `skipped_constraints` entry naming a constraint that did not exist. That is a false undetermined, which is the mirror image of the false clean this PR exists to prevent. A verdict has three answers, and the third one has to mean something. If it fires on a shape that was fully evaluated it stops carrying information, and the next person to see `null` on a clean run learns to ignore it.

Your drift-proof argument survives the fix intact, which is why I want the fix rather than the exception. Inspect `?o` for those two predicates only and skip when the value is anything other than `false`. Every other `sh:` predicate keeps the name-only rule, so a construct someone adds next year still lands in `skipped` without anyone remembering to update a list.

Some detail to hold while you write it:

1. Match on the boolean value, not the lexical form. `false` and `"false"^^xsd:boolean` are the same term in RDF 1.1, but `"false"` as a plain literal is not, and it is not a legal value for either predicate.
2. A non-boolean value on `sh:closed` or `sh:deactivated` should stay in `skipped`. That is a shape the validator cannot honour either way, and silently reading it as false would be a guess.
3. `sh:deactivated true` stays skipped for the reason you already gave, which is the right call and worth keeping in the comment: SHACL says a deactivated shape must not be evaluated, this validator evaluates it, so the predicate is not implemented.

Worth a test beside the three you have on decision 2: a shapes graph carrying `sh:closed false` and `sh:deactivated false` keeps a boolean verdict and no `skipped_constraints` key, mutation-checked by reverting to the name-only filter.

One thing I checked and it is fine, so you do not need to look at it. A predicate that is whitelisted but not actually implemented would be precisely the false clean, so I went through `sh:sparql` before believing the list. The validator really does read `sh:select` and execute it (`src/shacl.rs:494-613`), including the blank-node focus node exclusion it reports as skipped, so whitelisting it is honest.

On sequencing. #117 is merged. This PR and #122 conflict in `CHANGELOG.md` only, one hunk, both appending under `[Unreleased]` / `### Fixed`; the `src/shacl.rs` hunks do not overlap at all. I am taking #122 first because it is the smaller surface and it has no open question on it, so this one will want a one-line changelog rebase along with the `?o` change.

Everything else here I am happy with, including both of the decisions you flagged as reversible. Restricting the complement to the `sh:` namespace is the right call for the reason you gave, that a predicate outside `sh:` on a shape node is an axiom or an annotation by definition, and binding the shape as a variable rather than splicing a blank node label into the query text is a bug you found and fixed before I saw it.

fabio-rovai:
Stop before you write this one. I took it, and I want you to know inside the hour rather than after you have spent an evening on it.

Your commit is cherry-picked unmodified into #123 with your authorship intact, the `?o` fix sits on top as a second commit, and the `CHANGELOG.md` conflict against #122 is resolved by keeping both entries. Nothing in your implementation is changed. Once #123 lands I will close this PR as merged-by-proxy rather than as rejected, because that is what happened to it.

This is not a comment on your turnaround, which is faster than mine. It is that I had the tree open, the fix is nine tokens of SPARQL, and asking you to rebase a 395-line PR across two merges for that is a worse trade than writing it myself.

Doing it turned up a correction and a decision, both yours to disagree with.

The first is a correction to what I told you three hours ago. I said the guard should be `isLiteral` then `datatype` then equality, so that a non-literal or non-boolean value raises no error and stays skipped. The reasoning was right and the claim about this codebase was wrong: with the guard removed, all four unreadable-control cases stay green, because oxigraph answers `"false"^^xsd:string = false` with false rather than the type error SPARQL 1.1 permits on RDFterm-equal. So the guard changes no answer here and no mutation reddens it.

I kept it anyway and wrote the measurement into the comment beside it, marked as deliberately unmutatable, because the failure it guards is the one this whole complement exists to prevent. An error inside `FILTER` drops the solution rather than failing loudly, so on an evaluator that does raise, the unguarded form stops skipping precisely the controls it cannot read. That is a false clean arriving through the code path written to stop false cleans, on a dependency bump nobody would think to test. If you would rather have the simpler query and take the risk, say so and I will cut it, but I would want the reason in the changelog rather than only in a diff.

The second is `sh:ignoredProperties`, which I left out of the whitelist on purpose and which I did not raise with you earlier. It modifies `sh:closed` rather than constraining anything by itself, so wherever it has any effect the shape also carries `sh:closed true`, which is already recorded and already suppresses the verdict. Carrying it without `sh:closed true` is a shape that asks for nothing. Whitelisting it would be equally defensible and slightly cleaner; leaving it out costs a redundant `skipped` row in a report that is null regardless. I went with the version that keeps the whitelist meaning "predicates this validator reads".

Everything else in your PR went in as written, including both decisions you flagged as reversible, and neither wanted reversing.

On what is next. #117 and #122 are both merged, so everything you sequenced behind them is unblocked. #118 is yours whenever you want it. The snapshot half of #108 is the larger one and it is still a design question rather than a patch, and you were right to keep asking for the shape before writing: `validate` is default-graph-only on the same mechanism `check_shapes` just stopped being, so instance data in named graphs gives `focus_nodes: 0` and a `nothing_matched` null rather than a false pass. You should have a proper answer to that rather than a guess, so give me a day.

fabio-rovai:
Landed as `8ce929d` via #123, with your commit cherry-picked unmodified and your authorship on it. Closing this one as merged by proxy, not as rejected: the work in it is on main.

All ten CI legs green on the merge, including both Windows legs and the Docker build. Thanks for splitting it from #122 the way you did. Two independent PRs sharing a file but not a function meant the only conflict between them was one changelog hunk, and that is the reason both halves of #108 went in the same day.

PR #133, opened 2026-09-13
Fix the sh:sparql false conformance and violation attribution, and certify every OWL-RL inference

### LINKED ISSUE #131: shacl: violation objects omit the source shape, so a report with one shape per rule cannot be attributed
## What happens

`open-ontologies shacl <shapes>` returns violation objects with exactly these keys: `constraint`, `focus_node`, `message`, `severity` (see the `json!` blocks in `src/shacl.rs`, for example around lines 672, 796, 903). Nothing names the shape that produced the violation. The W3C SHACL validation report vocabulary carries `sh:sourceShape` (and `sh:sourceConstraintComponent`) on every `sh:ValidationResult` precisely so a consumer can tell which rule fired; pyshacl emits both.

## Why it matters

The usual pattern for a findings table is one shape per defect class, so that the validation report *is* the table. With this output the only way to attribute a violation to its shape is to embed an identifier in `sh:message` and parse it back out of free text. That is what I had to do in `sociotechnical-matrix-assurance/pipeline/verdicts.py` (39 shapes, one per matrix cell): the harness recovers the cell id with a regex over `message`. Any shape without a hand-written distinguishing message is unattributable.

## Reproduce

```
printf 'load ontology/stm.ttl\nload data/matrix.ttl\nload data/records.ttl\nshacl shapes/cell_shapes.ttl\n' \
  | open-ontologies --no-connect --data-dir /tmp/oo batch -
```
with the files from https://github.com/fabio-rovai/sociotechnical-matrix-assurance (39 `sh:NodeShape`s each carrying one `sh:sparql` constraint and a `stm:forCell` annotation). Every violation looks like:

```
{"constraint":"sparql","focus_node":".../record-0000000","message":"Cell DEV-D7 is not fully evidenced: ...","severity":"Violation"}
```

## Suggested fix

Add `source_shape` (the shape IRI) and `source_constraint_component` (for example `sh:SPARQLConstraintComponent`, `sh:MinCountConstraintComponent`) to every violation object, and `result_path` where a property shape is involved. Keep the existing keys unchanged so current consumers keep working.

## Related, already handled on my side but worth a note in the docs

The CLI store is in-memory per process, so `load` followed by `shacl` as separate commands validates an empty store. The engine reports this correctly (`conforms: null`, `focus_nodes: 0`, `unmatched_shapes` listing every target) and that is good behaviour. The docs could say in one line that multi-step work must go through `batch` (or the daemon), because a harness that only reads `violation_count` will read the empty-store case as "no violations".

Engine version: 1.3.0 (`status` output), HEAD 26f2672.

### LINKED ISSUE #132: shacl: sh:sparql constraints report zero violations when any focus node conforms (false conformance)
## Summary

A `sh:sparql` constraint with `FILTER NOT EXISTS` is evaluated as if the question were "does every focus node violate", not "which focus nodes violate". When at least one focus node conforms, the engine reports `conforms: true` and zero violations for that shape, even though other focus nodes violate it. This is a soundness bug: a data set with one clean record hides every dirty one.

## Minimal reproduction

data.ttl
```
@prefix ex: <http://example.org/> .
ex:r1 a ex:Record ; ex:has ex:e1 .
ex:e1 ex:field ex:F ; ex:status "present" .
ex:r2 a ex:Record ; ex:has ex:e2 .
ex:e2 ex:field ex:F ; ex:status "absent" .
ex:r3 a ex:Record .
```
shapes.ttl
```
@prefix sh: <http://www.w3.org/ns/shacl#> . @prefix ex: <http://example.org/> .
ex:NeedsF a sh:NodeShape ; sh:targetClass ex:Record ;
  sh:sparql [ sh:message "record lacks a present F" ;
    sh:select """SELECT $this WHERE { FILTER NOT EXISTS { $this <http://example.org/has> ?e . ?e <http://example.org/field> <http://example.org/F> ; <http://example.org/status> "present" . } }""" ] .
```
```
printf 'load data.ttl\nshacl shapes.ttl\n' | open-ontologies --no-connect --data-dir /tmp/store batch -
```
Engine: `conforms: true, violation_count: 0`.
pyshacl 0.40.1 on the same files: `conforms: False`, focus nodes `ex:r2`, `ex:r3`.

Change `ex:e1`'s status to `"absent"` so that all three records fail: the engine now reports `violation_count: 3`. So per-focus-node evaluation works only in the degenerate case where every focus node fails.

## Where it bit

`sociotechnical-matrix-assurance` (39 shapes of this form, 13 records): Python set-based and pyshacl agree on 245 violating (record, cell) pairs; the engine reports 52, which is exactly the 4 cells that fail for all 13 records, 13 times each, and nothing for the other 193 pairs.

## Expected

One `sh:ValidationResult` per focus node for which the SELECT returns at least one solution with `$this` bound to that node, per SHACL-SPARQL section 5.3. Pre-binding `$this` per focus node (or evaluating once with `$this` free and grouping results by the bound value) both give the right answer; what is happening now looks like a single evaluation whose emptiness is tested rather than whose bindings are read.

Engine 1.3.0, HEAD 26f2672. Related: #131 (violations carry no source shape), #98 (sh:sparql silently skipped, closed).

fabio-rovai:
Closes #132. Closes #131.

## The two issues

**#132** was a soundness bug. A `sh:sparql` constraint was evaluated by wrapping the author's `SELECT` as a subquery under a `VALUES ?this { ... }` clause. SPARQL evaluates a subquery bottom-up with no outer variable in scope, so `$this` was unbound inside it: `FILTER NOT EXISTS { $this ... }` asked whether ANY node matched rather than whether this one did. One clean record hid every dirty one, and the degenerate case where every record failed looked correct, which is how it survived. `$this` is now pre-bound per focus node through SPARQL substitution, which is the mechanism SHACL-SPARQL section 5.3.2 specifies. Blank-node focus nodes are evaluated instead of excluded, and a bound `?message`, `?path` or `?value` reaches the result.

On the 39-shape corpus the issue was filed from, this engine and pyshacl 0.40.1 now return the same answer: 249 results over 245 distinct focus-node and shape pairs.

**#131** asked for `sh:sourceShape` on every result. Every violation now carries `source_shape`, `source_constraint_component` and `result_path`, with the existing keys unchanged. `source_shape` names the shape that carries the constraint, so for a constraint under `sh:property` it is the property shape and not the node shape containing it. That distinction is the point of the field: two property shapes on one path under one node shape are the case the issue describes, and naming the node shape for both leaves them as indistinguishable as they were. pyshacl returns the property shape and this agrees with it. The enclosing node shape is still reported, under `node_shape`.

The issue also asked for a one-line note that multi-step CLI work must go through `batch`, because the store is in-memory per process. That is in `docs/lean-certificates.md`.

## Why the branch is larger than two fixes

Fixing #132 meant reading every rule in the forward-chaining reasoner and asking what semantic condition would make it sound. Two had none. `cls-svf1` derived `x rdf:type C` from `C rdfs:subClassOf ∃p.D`, which is the converse of the axiom, and it treated a class IRI in object position as an instance of itself. Both had been there since the extended profile was written, and no test caught them because a test catches the bugs someone thought of.

So the branch adds a check that does not depend on the author having thought of the bug. `reason --certificate DIR` writes every inferred triple with the rule that produced it and the premises the rule read. `lean/` holds a checker for that record whose soundness is a machine-checked theorem, `OOCert.certificate_sound`, in core Lean with no Mathlib and no `sorry`. A certificate it accepts contains only triples entailed by the asserted graph.

`lean/OOCert/Witness.lean` shows the theorem is not vacuous, which matters because a soundness result about an unsatisfiable semantics proves nothing. It exhibits a model of an arbitrary graph, exhibits a triple that is not entailed, and proves that the derivation removed from `cls-svf1` is refuted by a model of its own premises while the half the reasoner still makes is entailed. That turns "we removed a rule we could not justify" into "we removed a rule that was unsound".

## What an audit of this branch then found

The branch was reviewed adversarially before this PR, with every finding independently refuted or confirmed and every confirmed one reproduced by hand against pyshacl. Fifteen survived. Two were defects in this branch's own work: the CI job that gates the proofs could not pass on a runner, and the `#131` attribution was wrong for the case `#131` was filed about. Both are fixed here.

Seven predate the branch, and three of those are worth calling out:

- `sh:prefixes` was never read and every `sh:declare` in the shapes graph was merged into one prologue. A prefix bound to two namespaces emitted two `PREFIX` lines, SPARQL takes the last, and which one won was decided by store row order. The constraint pointing at the loser matched nothing and the run reported `conforms: true` with nothing recorded. Two shapes files that were the same RDF graph could give opposite verdicts. Declarations are now scoped per constraint as SHACL 5.2.1 says, and an ambiguous merge is refused with a recorded reason rather than guessed.
- The helper that strips a literal's quoting searched the whole string for `^^` and `"@` and truncated there, so `FILTER(?d < "2025-01-01"^^xsd:date)` was chopped mid-query. The constraint then failed to parse and was recorded as unrunnable with an error blaming the author, leaving `conforms: null` and exit 0. Date and numeric comparisons are the most common SPARQL constraints anyone writes.
- `prp-symp`, `prp-inv1`, `prp-inv2` and `eq-sym` take their conclusion's subject from an object position, so a literal object produced a triple no serialisation can express. Materialisation then failed on the whole batch and lost every inference from that run, at a line number that moved between runs because it indexes a serialisation ordered by hash iteration. That non-determinism is what made it look like malformed input.

The rest are in the CHANGELOG: a blank-node node shape acting as a SPARQL wildcard, an ignored `sh:deactivated`, a reasoner that was not a fixpoint of its own rule set, a `cls-hv1` that was not W3C's `cls-hv1`, an `rdfs3` guard coarser than its invariant, and a checker exit code that did not match its documented contract.

One fix was withdrawn after measurement. Deduplicating identical validation results looked right for a node selected by two target declarations, but SHACL emits one result per SPARQL solution and pyshacl does too, so deduplicating broke the exact agreement above. The double-counting stays, documented.

## What the Lean layer does and does not guarantee

`OOCert.certificate_sound` says that if the checker accepts a certificate, every conclusion in it is entailed by the graph in `asserted.tsv`, under a model-theoretic semantics in which a term is a string and each semantic condition is the weakest the rules need. Weaker conditions admit more interpretations, so the result carries to the W3C RDF-based semantics. The semantics has no notion of datatypes or literal values, so the theorem says nothing about literals beyond their spelling.

The guarantee is relative to `asserted.tsv`, which is whatever the run started from: the store flattened across named graphs, in the store's post-parse canonical spelling rather than the source file's. The checker is sound and not complete, so it rejects entailed inferences it cannot pattern-match, and a rejection is a false alarm and never a false pass. The rule table is this engine's, not W3C's, so a `by_rule` histogram describes what this engine did and is not an OWL 2 RL conformance claim. Nothing outside the forward-chaining family is covered: not the SHACL validator, where the pyshacl differential remains the gate, not the SHIQ tableaux reasoner, and not the parsers.

## Verification

| check | result |
| --- | --- |
| test suite | 976 passed, 0 failed, 0 fixture skips |
| clippy, all targets | clean |
| Lean build, which is the proof check | green, axioms pinned by `#guard_msgs` |
| certificate corpus | 122 files, 37,133 derivations, 6 exclusions each named with a reason |
| pyshacl differential | no false clean, no false alarm |

The corpus is enumerated from `git ls-files`, so "every RDF file in this repository" means the same thing on every machine. It used to be five hand-named directories, three of which hold no RDF, so 46% of the repository's RDF was never walked and was excluded without being named. Widening it is what surfaced the literal-subject defect, which lived in `benchmark/`.

The `lean` CI job has never run on a runner. This PR is the first time it does.


fabio-rovai:
Superseded by #134, which is built on this branch: `git merge-base --is-ancestor lean-certificate-layer owl-rl-completeness` returns true, so every commit here is contained there. #134 carries the same `Closes #131` and `Closes #132`, plus the Lean certificate work those fixes led to. Closing this one so there is a single thing to review and merge.

PR #167, opened 2026-09-15
tcb: name the Lean kernel, the toolchain and the axiom pin, and test whether anything outside Lean can re-check them

fabio-rovai:
**Stacked on `shrink-tcb`, which is the base of this branch.** It rewrote
`docs/trusted-computing-base.md`, and this builds on that version.

A documentation defect that was certainly real, and an investigation whose answer was not known in
advance.

## 1. The TCB page omitted its own foundation

`docs/trusted-computing-base.md` ended with a list of what is still trusted: `oxrdf`'s escaping,
`oxiri`'s IRI validation, premise ordering, the reasoner's completeness, the SHIQ tableaux reasoner,
the Rust SHACL validator, fifty thousand lines of `src/`. It did not list the Lean kernel, the
toolchain, or the axioms. That is the largest trusted component in the system, absent from the one
document whose job is to enumerate what is trusted.

Entries 7, 8 and 9 repair that. Entry 7 is the kernel's soundness and why v4.33.1 is not an
arbitrary pin. Entry 8 is the elaborator, what a green `lake build` actually asserts, and the fact
that `lean-toolchain` pins a version and not any bytes. Entry 9 is the axiom footprint.

**The axiom pin is ENFORCED, not conventional.** All 121 `#print axioms` lines in `lean/` sit under
`#guard_msgs`, so a drifted footprint is a build error rather than a message nobody reads.

## 2. `docs/independent-rechecking.md`

Isabelle bought independence of the SPEC READING. It bought nothing about the KERNEL, and the two
get conflated. This is the investigation of whether a Lean proof can be re-checked by something that
is not Lean.

It can, and it was. `nanoda` 0.4.17, a Lean 4 kernel written in Rust sharing no code with Lean,
accepted this repository's export, exit 0. Dedukti cannot do it and the reasons are structural.
The recommendation is to add `leanchecker` to CI and do nothing else yet, with three named triggers.

## Which claims are verified by running, and which are from documentation

The new document marks every claim **RAN** or **READ** in place. Summarised:

### VERIFIED BY RUNNING on this machine, 15 September 2026, Lean v4.33.1

| claim | evidence |
|---|---|
| `lake build` in `lean/` is green | 105 jobs, no warnings, exit 0 |
| all 121 `#print axioms` sit under `#guard_msgs` | script over `lean/`, zero unguarded |
| a mismatched pin FAILS the build | `lean` exit 1 on a deliberately wrong pin |
| a `sorry` UNDER a pin FAILS | exit 1, footprint reported as `[sorryAx]` |
| a `native_decide` under a pin FAILS | exit 1 |
| a `sorry` with NO pin above it does NOT fail | warning only, **exit 0** |
| `native_decide` yields `._native.native_decide.ax_1_1`, not `Lean.ofReduceBool` | printed footprint on v4.33.1 |
| no grep/test/CI step bans `sorry` or `native_decide` | grep over `tests/`, `.github/`, `scripts/` |
| `lean/` has no `axiom`, `unsafe`, `partial`, `@[implemented_by]`, `@[extern]` | grep |
| `lake-manifest.json` lists zero packages | file |
| **44 hand-written theorems are reached by no pin** | Lean metaprogram over all 42 modules; 506 of 1,178 theorem constants covered |
| **0 theorems use an axiom outside the three** | same metaprogram, all 1,178 |
| `OOCert.lean` does not import `Refute`/`RefuteWitness` | file; cost 13 pins on the first pass |
| Lean v4.33.1 has no export flag | `lake env lean --help` |
| `lean4export` builds at v4.33.1 | tag `v4.33.0` + toolchain override, `lake build` 6 jobs, exit 0 |
| export of `OOCert Shacl.All Dl Fol` succeeds | 515 MB, 9,850,420 lines, 97 s, exit 0 |
| `nanoda` 0.4.17 builds | `cargo build --release`, 74 s |
| **`nanoda` ACCEPTS this repository's export** | exit 0, 25 min; skipped-axiom line names `sorryAx` and `Lean.ofReduceBool` as declared-but-unused |
| `nanoda` first run FAILED | panic at `src/tc.rs:230`, `Lean.trustCompiler` not permitted |
| `leanchecker` ships in the v4.33.1 toolchain and PASSES | exit 0 on all four roots via `lake env` |
| bare `leanchecker` fails with `incompatible header` | resolves the default toolchain |
| Mathlib's nanoda job allows exactly those four axioms | fetched `mathlib4/.github/workflows/daily.yml` |
| Mathlib CI pins nanoda branch `debug`, last commit 2025-10-26 | same file, plus `git log origin/debug` |
| `lean2dk` pins v4.18.0-rc1 / v4.22.0-rc4, `Lean4Less` v4.28.0-rc1 | fetched each branch's `lean-toolchain` |
| the arena's `checkers/` directory holds 24 entries | GitHub contents API |
| `lean4export` NDJSON transition landed 2025-12-13 | `git log` on the repo |

### FROM DOCUMENTATION, not executed here

- The Lean kernel soundness bugs fixed in v4.33.1, and that the arena scores the v4.28.0 kernel
  wrong on six tests. From de Moura's post-mortem and the arena's published results.
- `leanchecker` "is not an external verifier, simply a tool to detect environment hacking". Its own
  source docstring.
- `lean4lean` "is not really an independent implementation", its `Verify` `sorry` count, and that it
  does not implement `reduceBool`. Its README, `divergences.md`, and the TYPES 2025 paper, DOI
  `10.4230/LIPIcs.TYPES.2025.2`.
- Everything about what `lean2dk` produces: the single module it checks end to end, the personal
  Dedukti fork it needs, the value-stubbing, the `STRLIT.FIXME` placeholder, the two-to-three orders
  of magnitude blow-up. Its README and the 2026 Paris-Saclay thesis. **Not run**, because the
  toolchain gap alone disqualifies it.
- Lambdapi's "rules are simply ignored" limitation on its Coq and Lean exports. Its `doc/cli.rst`.
- Logipedia's Lean 3.4.2 backend and its dead sites.
- Dedukti's commit-volume decline and the EuroProofNet funding position.

## Verification

- `cargo clippy --all-targets -- -D warnings`: **exit 0, zero warnings**.
- Full `cargo test --all-targets`: complete result in the thread, not truncated.
- All 29 `TCB-N` identifiers are preserved, so
  `the_trusted_base_document_names_every_property_tested_here` still holds.

## What this does NOT do

- **No Lean proof was modified.** `git diff shrink-tcb --name-only -- lean/` is empty.
- **No dependency was added.** `lean4export`, `nanoda` and `leanchecker` were run out of tree.
  Nothing in `Cargo.toml`, `lakefile.toml` or CI references them.
- It does not close the 44-theorem pin gap. That is named with its fix and left as work.
- It does not add `leanchecker` to CI. That is the recommendation, deliberately left as a separate
  decision rather than smuggled in with a documentation change.

## One thing to push back on if you disagree

Three comments in this repository claim a `sorry` "anywhere" fails the build: the `lean` job in
`.github/workflows/ci.yml`, `lean/Dl/All.lean` and `lean/Fol/All.lean`. They are wrong. I left them
in place and named them in entry 9 rather than correcting them, on the reasoning that fixing the
prose without closing the hole makes the gap harder to find. If you would rather have the comments
corrected now, say so.


fabio-rovai:
## Verification, complete and not truncated

`cargo test --all-targets`, run to completion on this branch:

```
exit code:                 0
test binaries reporting:   130
binaries FAILED:           0
individual test failures:  0
panics or compile errors:  0

aggregate: 1312 passed; 0 failed; 1 ignored; 0 measured
```

The Lean-dependent binaries ran for real rather than skipping, which matters because
they skip silently when `lake` is absent:

| binary | result |
|---|---|
| `lean_certificate_test` | 6 passed, 110.51s |
| `cross_kernel_differential_test` | 15 passed, 217.53s |
| `dl_model_certificate_test` | 12 passed |
| `lean_refutation_test` | 9 passed |
| `lean_fol_model_test` | 9 passed |
| `lean_horn_certificate_test` | 5 passed |
| `shacl_core_verified_test` | 4 passed |

`cargo clippy --all-targets -- -D warnings`: **exit 0**, zero warnings, zero errors.

`lake build` in `lean/`: **exit 0**, 105 jobs, no warnings. No Lean file was touched
on this branch (`git diff shrink-tcb --name-only -- lean/` is empty), so that is the
same proof check as on the base.

### One process note

The first two attempts at the full suite were killed part-way through the dependency
build, not by a test failure. The machine was at load average 26 with other work on it.
The passing run above is a complete one: `cargo test --all-targets --no-run` to build,
then `cargo test --all-targets` to run, both to exit 0.

PR #172, opened 2026-09-15
main is red: the corpus figures are stale and every PR inherits it

fabio-rovai:
`the_corpus_exercises_prefix_visibility_at_depth` **fails on `main`**. The corpus is 2,405
certificates with 505 exercising prefix visibility; `README.md` and
`docs/reasoning-systems-inventory.md` still say 2,075 and 484.

Every open PR inherits `main`, so **all thirteen were failing the same required check at once.**
That is why the merge queue stopped moving while appearing to be merely waiting for CI.

## What this exposes

- A red `main` can sit unnoticed while attention is on the pull requests.
- `gh pr checks` reports a **cancelled, superseded** run as `fail`, which is indistinguishable at a
  glance from a real failure. Several hours of today's red marks were cancellations.

## The fix

The figures are derived by the test from the corpus it generates, so the documents are corrected to
match the run rather than the test relaxed to match the documents. That is what the test's own
failure message instructs.

Verified locally: gate passes.


fabio-rovai:
Closing: I had this backwards.

`main` was **not** red. The corpus gate computes 2,075 on a clean checkout, which is what `main`'s documents say and what CI measures. My machine computed 2,405 because `demo/derived/` holds 19 generated `.ttl` files from an earlier demo run. That directory is gitignored, so CI never sees them, but `source_graphs()` in the differential test scans it anyway.

So this PR would have broken `main` by replacing correct figures with ones true only on a machine that had run the demo.

The real defect is that the corpus scans a **generated** directory, which makes the measurement depend on local state. Filing that separately.

PR #221, opened 2026-09-19
ci: build hnswlib without -march=native so the cached wheel runs on every runner

fabio-rovai:
The `python (open-ontologies-lite)` job fails on main about every other run with:

```
Fatal Python error: Illegal instruction
  File ".../python/tests/test_align.py", line 7 in <module>
```

Run history on main today, no Python change between them:

| run | sha | python job |
|---|---|---|
| 20:16 | af31f20 | failure |
| 20:06 | b583d62 | success |
| 19:14 | 9179bc9 | failure |
| 18:59 | fb707af | success |

Cause: hnswlib 0.8.0 ships only `hnswlib-0.8.0.tar.gz`, and its `setup.py` adds `-march=native` unless `HNSWLIB_NO_NATIVE` is set. uv caches the wheel it built. The hosted runners are not one CPU, so a wheel built on a machine with AVX-512 gets restored on one without it and the import traps.

Fix: `HNSWLIB_NO_NATIVE=1` on the install step, and a `cache-suffix` on `setup-uv` so the native wheel already in the cache is never restored again.

Also affects #219 and #220, which show the same failure and will go green on this leg once this is merged and they are rebased.


fabio-rovai:
Superseded by #223, which is on `main`.

#221 set `HNSWLIB_NO_NATIVE=1`, which is necessary and not sufficient: `setup-uv` caches BUILT WHEELS across runners, so once any runner compiled hnswlib with `-march=native` the cached artifact was restored onto machines that could not execute it, and the env var on a later run changed nothing. `--reinstall-package` and `--refresh-package` do not bypass that cache either (measured: "Prepared 1 package in 0.39ms"). Only `--no-cache` forces the compile, which is what #223 does, together with this env var and a subprocess probe in `test_align.py` that names which import dies rather than letting `importorskip` hide it.

Closing rather than rebasing: the branch's change is already in main as part of the larger fix.

PR #253, opened 2026-09-21
Refuse a release tag that does not carry the version it claims

fabio-rovai:
`open-ontologies status` on a binary built from tag `v1.6.0` reports `"version":"1.5.0"`.

The tag and the `version` field in `Cargo.toml` are set by separate commits, so a tag can land before the bump. When it does, the release builds and publishes cleanly while the binary reports the previous version. `status` is the only version the engine reports about itself, so anything recording which engine produced an artifact records the wrong one, and a consumer pinning a release and verifying by the reported version cannot tell a correct build from a failed pin. The failure is silent and it flatters the release: green CI, artifacts present, only the number wrong.

## The current state

`scripts/check-release-version.sh --all` on this repo:

```
v0.1.1: Cargo.toml says 0.1.0
v0.1.2: Cargo.toml says 0.1.0
v0.1.3: Cargo.toml says 0.1.0
v0.1.4: Cargo.toml says 0.1.0
v0.1.5: Cargo.toml says 0.1.0
v0.1.6: Cargo.toml says 0.1.0
v0.1.8: Cargo.toml says 0.1.0
v0.1.9: Cargo.toml says 0.1.0
v1.2.1: Cargo.toml says 1.2.0
v1.4.0: Cargo.toml says 1.3.0
v1.6.0: Cargo.toml says 1.5.0

11 tag(s) do not carry the version they claim.
```

`v1.6.0` is the current release.

Reproduce without building anything:

```
git worktree add /tmp/oo-v160 v1.6.0
cd /tmp/oo-v160 && CARGO_TARGET_DIR=/tmp/oo-v160-target cargo build --release --bin open-ontologies
/tmp/oo-v160-target/release/open-ontologies status
# {"status":"ok","triples_loaded":0,"version":"1.5.0"}
```

## What this adds

`scripts/check-release-version.sh <tag>` compares the tag against `Cargo.toml`, and `--all` audits every `v*` tag so the historical cases stay visible rather than only new ones being prevented.

The release workflow runs it as a `verify-version` job that `build` depends on, so a mismatched tag fails before any artifact carrying the wrong number exists. The docker workflow runs it as a step guarded by `startsWith(github.ref, 'refs/tags/v')`, since that workflow also runs on `main` and on pull requests where there is no tag to check.

Verified in all three directions: it passes on a matching tag, fails on a mismatched one with the fix spelled out, and the `--all` audit reproduces the table above.

`python-release.yml` is untouched. It triggers on `py-v*` and tracks the Python package version, which is a separate scheme.

## Note on the existing tags

This only prevents new occurrences. The eleven above are already published, and `v1.6.0` in particular will keep reporting `1.5.0` unless it is retagged, which is a separate decision.


fabio-rovai:
Verified here and took one thing further.

The central claim reproduces exactly: `--all` finds the same 11 tags, including `v1.6.0` reporting 1.5.0. The gate fails on a mismatch and passes on a match, and the single-tag mode is right for the workflow because `actions/checkout` on a tag push checks out the tag, so the tree it compares against *is* the tag's tree.

One gap. The body says the version is checked "before anything is built, so no artifact carrying the wrong number is ever produced". `build` needed `verify-version`; **`checkers` did not**. On a tag already known to be wrong the whole Lean development still compiled, about forty minutes of runner time. Nothing was ever *published* without the gate, since `release` needs `build`, so the property that matters held. The sentence did not. `checkers` now has the same one-line gate.

Worth recording alongside: #248 fixed main going forward by bumping `Cargo.toml` to 1.6.0. This stops the same thing recurring at tag time. The two are the halves of one defect, and neither fixes the released v1.6.0 binary, which still reports 1.5.0 and will until there is a v1.7.0.

